PR #13, opened 2021-02-05
Pokemon Battle Engine

### LINKED ISSUE #11: Add the Agent interface
We need an interface to represent the entities that engage in a battle. The idea is to have an `Agent` interface that, when implemented, allows an end user or bot to receive information about the battle and complete a turn. 

See: https://github.com/StevensSEC/pokemonbattlelib/wiki/Requirements#before-the-battle

### Tasks

- [ ] Define the functions of `type Agent interface` that are necessary in order to accommodate the 2 use cases listed above.

adapap:
The engine will use an observer pattern to dispatch event hooks so that clients can write code to interface with the library in a certain state (`onBeforeBattle`, `onAfterPlayer1Turn`, etc). Basic configuration will be done with simple methods on the engine, such as `engine.AddPokemon(...)`. For the user input, still deciding on using channels (state will be emitted and the user will have to write logic to handle the current state) or have the user manually call methods to advance the game.

## Channel/Emitting System
```golang
// client, concurrent
for {
  select {
    case <- "player1_turn":
      // make an action
      e.Attack(move)
    case <- "battle_end":
      fmt.Println("Winner:", e.BattleResult())
  }
}
```

## Manual State Tracking
```golang
// client, synchronous
for {
  if e.state == PLAYER1_TURN {
    e.Attack(move)
  } else if e.state == BATTLE_END {
    fmt.Println("Winner:", e.BattleResult())
  }
}
```

Closes #11


dyc3 (commented):
It's an interesting approach, but I think it's a little too complex. There's definitely no need for an `Engine` struct, as the `Battle` struct will hold that information.

Another problem I see is that there is only one listener for the events. I think that this makes too many assumptions about how the user will use the library, because both participants in the battle could have wildly different implementations, (eg. one could be a human via a web interface, and the other could be a bot that runs on the server). It also makes the user do more work to provide information to the participants.

(nit) I also don't think its a good idea to have all those constants as strings. It's kinda just wasted space, especially because we can just implement the `Stringer` interface

I think we should try something different.

BTW, #11 only calls for the outlining what the interface should look like. Don't bother with implementing the methods for now.

client_test.go
```diff
@@ -0,0 +1,13 @@
+package pokemonbattlelib
+
+func handleTurn() interface{} {
+	return "123"
+}
+
+func ExampleClient() {
```
dyc3, line 7:
```suggestion
func TestExampleClient(t *testing.T) {
```

battle_test.go
```diff
@@ -0,0 +1,27 @@
+package pokemonbattlelib
+
+import "fmt"
+
+func ExampleBattle() {
```
dyc3, line 5:
```suggestion
func TestExampleBattle(t *testing.T) {
```

client.go
```diff
@@ -0,0 +1,58 @@
+package pokemonbattlelib
+
+import (
+	"fmt"
+	"time"
+)
+
+type Client struct {
```
dyc3, line 8: (nit) Agent is more in line with the vocabulary used in game AI

battle.go
```diff
@@ -1,5 +1,117 @@
 package pokemonbattlelib
 
+import "fmt"
+
 type Battle struct {
 	// TODO
+	State   BattleState
+	Player1 *Client
+	Player2 *Client
```
dyc3, line 9: Let's just make this an array.

battle.go
```diff
@@ -1,5 +1,117 @@
...
+}
+
+// Swaps player 1 with player 2 in a battle
+func (b *Battle) SwapPlayers() error {
+	if b.Player1 == nil || b.Player2 == nil {
+		return fmt.Errorf("battle does not have two clients/players")
+	}
+	b.Player1, b.Player2 = b.Player2, b.Player1
+	return nil
+}
```
dyc3, line 71: Is this function necessary?

dyc3 (changes requested):
This will work for a first draft. Just fix up some of the little things

Is there any particular reason you didn't use an interface to interact with the agents?

adapap:
Interface doesn't let you keep track of any state, so you'd need either getters/setters, another struct, or global variables. Also, you have to type check an interface in some cases whereas struct is usually implicit

battle.go
adapap: Only thought this was needed in case the player wanted to reverse the first opponent to act and not deal with the battle fields manually. I guess if it becomes an array then i'm not sure?

dyc3:
Also, it looks like one of the tests is failing.

dyc3 (approved):
The code coverage is a little rough, but it's probably fine.

Issue #10, opened 2021-01-25
Add Move type


carpenterd777:
We need a type to represent the metadata of a Pokemon's moves at runtime. This won't be responsible for handling side effects of moves yet.

See: https://github.com/StevensSEC/pokemonbattlelib/wiki/Requirements#moves

### Tasks

- [ ] Define the contents of `type Move struct` that are necessary to have in a Pokemon battle, or to be able to obtain external metadata.
- [ ] Implement the `Stringer` interface for `Move`. Example: https://tour.golang.org/methods/17

Issue #6, opened 2021-01-24
Add the Pokemon Type


dyc3:
Obviously, we need a type to represent a Pokemon in a battle. Keep memory footprint in mind when considering what types to use. Feel free to add any enums that you need to accomplish this.

See: https://github.com/StevensSEC/pokemonbattlelib/wiki/Requirements#pokemon

It's not necessary for the first iteration of this type to be perfect. We can change it later if needed.

### Tasks

- [x] Define the contents of `type Pokemon struct` that are necessary to have in a Pokemon battle, or to be able to obtain external metadata. 
    - For example: the Pokemon's name should not be stored in the struct, but the National Pokedex number should since it can be used to look up the Pokemon's name.
- [x] Implement the `Stringer` interface for `Pokemon`. Example: https://tour.golang.org/methods/17

PR #15, opened 2021-02-06
Codegen for pokemon

### LINKED ISSUE #5: Data entry and accessing it at runtime
We need to be able to access metadata about pokemon, moves, etc at runtime. `pokedex.go` should manage this information.

### Desired Solution

Code generation based on files. When `go build` is run, it should run a script to generate the code with the metadata. This is accomplished with the magic comment `go:generate`.

This example will run `gen.go` before building the rest of the package.
```go
//go:generate go run gen.go
```

The script should read some files out of the `data/` directory, and generate code based on it's contents.

### Tasks

- [x] File format(s) or schemas for metadata
- [x] API for accessing metadata
- [x] Script for generating code
- [x] Data entry
    - [x] Pokemon
    - [x] Moves

dyc3:
- add code generation script for pokemon data
- add a method to get pokemon by national dex number

Closes #5

I'm unable to add complete codegen for moves, as I'm waiting for #10. I'm also unable to add most of the fields for pokemon because I'm waiting for #6.

The generated code is checked into the repo because it requires `go generate` to generate, and does not get automatically generated with `go build`.


data/abilities.csv
```diff
@@ -0,0 +1,294 @@
...
+115,ice-body,4,1
+116,solid-rock,4,1
+117,snow-warning,4,1
+118,honey-gather,4,1
+119,frisk,4,1
+120,reckless,4,1
+121,multitype,4,1
+122,flower-gift,4,1
+123,bad-dreams,4,1
+124,pickpocket,5,1
```
carpenterd777, line 125: I'm noticing some Abilities from Gen V over here, do we want to include these?

data/conquest_episode_names.csv
```diff
@@ -0,0 +1,77 @@
+episode_id,local_language_id,name
```
carpenterd777, line 1: I see that there's a lot of data in here, I think these are like the chapter names for the Conquest spin-off games, do we need all of this extra data?

dyc3: I just included all of it because it was kinda hard to determine what was needed and what wasn't necessary.

data/abilities.csv
dyc3: I'm excluding them in the code generation.

data/gen.go
```diff
@@ -0,0 +1,301 @@
...
+	// get all valid version ids
+	valid_version_groups := []int{}
+	version_groups_csv := getCsvReader("data/version_groups.csv")
+	for {
+		record, err := version_groups_csv.Read()
+		if err == io.EOF {
+			break
+		}
+		gen_id, err := strconv.Atoi(record[2])
+		if gen_id <= 4 {
```
carpenterd777, line 96: maybe we could have the `4` here be a constant, say `MOST_CURRENT_SUPPORTED_GEN`, at the top of the file so that if the library ever expands to support later generations the change will be easier to make

pokedex_GEN.go
```diff
@@ -0,0 +1,1010 @@
...
+	{ NatDex: 485 },
+	{ NatDex: 486 },
+	{ NatDex: 487 },
+	{ NatDex: 488 },
+	{ NatDex: 489 },
+	{ NatDex: 490 },
+	{ NatDex: 491 },
+	{ NatDex: 492 },
+	{ NatDex: 493 },
+	{ NatDex: 0 },
```
carpenterd777, line 500: do you know what caused all of these `0` entries to be created?

dyc3: Other forms of pokemon like deoxys and rotom are listed as separate pokemon

data/conquest_episode_names.csv
carpenterd777: gotcha, maybe we can open a new issue to prune some of this data later then

pokedex_GEN.go
carpenterd777: oh, right. do you foresee this causing any issues with our implementation of getting pokemon names? (the one in `pokemon.go`)

dyc3: Not really. We can worry about pokemon forms later.

carpenterd777 (approved):

Issue #8, opened 2021-01-24
Add the ElementalType type


dyc3:
We need a way keep track of single and dual type Pokemon.

It's not necessary for the first iteration of this type to be perfect. We can change it later if needed.

### Tasks

- [ ] Define a new type called `ElementalType`
- [ ] Define a utility method to determine the effectiveness of a Move/Pokemon type matchup. It should take the attacking move type and defending Pokemon type, and return some representation of it's effectiveness.
    - See type effectiveness chart: https://github.com/StevensSEC/pokemonbattlelib/wiki/Requirements-(Miscellaneous)#type-effectiveness-chart

dyc3:
TODO: assign @krixstin

PR #17, opened 2021-02-07
Added move type

### LINKED ISSUE #10: Add Move type

Jordan883:
closes #10


move.go
```diff
@@ -0,0 +1,20 @@
...
+	Type string
+	Category string
+	Max_PP int
+	Priority int
+	Power int
+	Accuracy int
+	//Side effects not implemented yet
+}
+
+//Stringer implementation
```
dyc3, line 17: (nit) obvious, should be omitted

move.go
```diff
@@ -0,0 +1,20 @@
...
+func (m Move) String() string{
+	return fmt.Sprintf("Name: %v\nType: %v\nCategory: %v\nMax PP: %v\nPriority: %v\nPower: %v\nAccuracy: %v\n", m.Name, m.Type, m.Category, m.Max_PP, m.Priority, m.Power, m.Accuracy);
```
dyc3, line 19: I feel like this is way too much information.

move.go
```diff
@@ -0,0 +1,20 @@
+package pokemonbattlelib
+
+import "fmt"
+
+//A type for moves
```
dyc3, line 5: Needs to be slightly more descriptive
```suggestion
// Represents a Pokemon's move. Moves can deal damage, heal the user or allies, or cause status effects.
```

dyc3 (changes requested):

move.go
Jordan883: Changed string to name, type, power, and acc in shorter format

move.go
Jordan883: Removed excess comments

dyc3 (approved):

carpenterd777:
This implementation has the `Type` field as a `string`. shouldn't it be an `ElementalType`?

dyc3:
That can be changed later, when #8 is done

PR #23, opened 2021-02-08
Hold Items

### LINKED ISSUE #20: Add the HoldItem type
Held items change the way that a Pokemon battle functions drastically. Each hold item belongs to a class (such as Stat Modifier, Medicine, Berries, or other effects. Representing items will be challenging, as there are many different types of items that produce different effects. Also, we must determine if we want to include all possible items or only those which have an affect on the battle itself.

This issue should be incremental, with the data representation being the first and simplest layer.

adapap:
Added all items into the generated output from `go generate`. This list is not filtered by anything, so over time we can reduce the items based on what is relevant. Some enums such as item categories and fling effects were also added.

Closes #20.


dyc3 (approved):
Try to avoid changing unrelated parts. It makes it significantly harder to review your changes.

carpenterd777 (approved):

PR #24, opened 2021-02-08
Removed extra csv files

carpenterd777:
- Removed pokemon conquest data
    - Pokemon conquest is a spinoff game that, to my knowledge, does not feature traditional Pokemon battles, so I didn't think we would need it here.
- Removed super contest data
    - Super contests were a minigame in Gen IV that expanded on the contests from Gen III.
- Removed region name data (Kanto, Johto, etc. did not seem relevant to battles)
- Removed species flavor text related files (Pokedex entries do not come up in battle)
- Removed pokemon shapes files (the shape of a Pokemon as define by the Pokedex)
- Removed habitat related files (where Pokemon can live)
- Removed pokeathlon stat data (a minigame added in HG/SS and not featured again)
- Removed berry firmness data (a nonfunctional attribute)
- Removed contest data (contests were a minigame from Gen III)
- Removed egg related data (eggs cant battle)
- Removed encounter-related data (data related to the various ways that Pokemon can be encountered, not the actual battle that results from one)
- Removed location-related data (city names, etc.)
- Removed pal park information (Pal Park was a way to transfer Pokemon from Gen III to Gen IV)
- Removed data about pokedexes (NOT dex numbers, data about the pokedexes themselves)
- Removed pokemon color information 

- Added list of used files in README


dyc3 (approved):
> eggs cant battle

Not with that attitude /s

PR #25, opened 2021-02-08
Update Battle Type

### LINKED ISSUE #7: Add the Battle type
We need a way to track the current state of a Pokemon battle.

It's not necessary for the first iteration of this type to be perfect. We can change it later if needed.

### Tasks

- [ ] Define the contents of `type Battle struct` that necessary for a Pokemon battle.

A-very-Cunning-ham:
Add basic fields for the Battle type. This may need to be updated later.

Closes #7

### DIFF
--- a/battle.go
+++ b/battle.go
@@ -5,12 +5,10 @@ import (
 )
 
 type Battle struct {
-	// TODO
-	State   BattleState
-	Players []*Agent
-
 	Weather  int  // one of the 6 in-battle weather conditions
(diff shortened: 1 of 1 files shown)


battle.go
```diff
@@ -8,7 +8,11 @@ type Battle struct {
 	// TODO
```
dyc3, line 8: Could you put them here instead, replacing the comment?

dyc3 (changes requested):

dyc3 (approved):

PR #26, opened 2021-02-08
Complete system to allow agents to make turns

### LINKED ISSUE #18: Mechanism to allow Agents to take turns
We need some way to process turns from `Agent`s, without having the agents directly modify the state of the battle.

dyc3:
- add Turn interface
- rework battles to be less confusing

closes #18

## Why a major rework was necessary

- Was hard to understand
- Could allow Agents to illegally modify the battle state
- Would be difficult to add double battles
- The event system did not make a lot of sense in the context of this library
- Made too many assumptions about how the library would be used
- Changing the order of turns depending on priority was non-trivial and unclear
- etc.

# How it works

To initialize a new battle
1. Agents get added to the Battle
2. Parties get added to the Battle
3. Parties get assigned to Agents
4. Agents get grouped into teams

To simulate the battle
1. call `Start()`
2. call `SimulateRound()` until the battle is complete

Currently, there is no way to know the result of a battle

The User has more control over when stuff happens, so it makes the event system redundant.

### Additional benefits

- User facing API is more clean
- Should be easier to write tests
- Agents become party agnostic
- Agents are now easier to manage


battle.go
```diff
@@ -1,115 +1,215 @@
...
+	"fmt"
 )
 
 type Battle struct {
-	Weather  int  // one of the 6 in-battle weather conditions
-	ShiftSet bool // shift or set battle style for NPC trainer battles
-	State    BattleState
-	Players  []*Agent
+	Weather      int  // one of the 6 in-battle weather conditions
+	ShiftSet     bool // shift or set battle style for NPC trainer battles
```
carpenterd777, line 9: I feel like this field is a little unclear (if this is `true`, is it shift or set?) maybe we can use an enum instead of a boolean? or rename it to `IsShift`

dyc3: That's not part of this PR, it was from #25. It can be changed later

carpenterd777: okay, gotcha

battle.go
```diff
@@ -1,115 +1,212 @@
...
+	b.agentParties[agentIdx] = partyIdx
+	return nil
+}
+
+// Set which agents have which allies. Not to be confused with `Party`.
+func (b *Battle) SetTeams(t [][]int) error {
+	// TODO: validate
+	b.teams = t
+	return nil
+}
```
carpenterd777, line 55: by naming this `SetTeams()` I imagine you had multi battles in mind (2 Trainers vs. 2 Trainers). do you think this will be able to cleanly support standard double battles (each trainer sends out 2 Pokemon from their party)?

dyc3: It should, because we can allow parties to have multiple active pokemon. I didn't implement that for the sake of keeping this PR as simple as possible.

carpenterd777 (approved):

battle.go
```diff
@@ -1,115 +1,212 @@
...
+
+	for _, apIdx := range turnOrder {
+		switch t := turns[apIdx].(type) {
+		case FightTurn:
+			fmt.Printf("TODO: Implement fight %v\n", t)
+		default:
+			panic("Unknown turn")
+		}
+	}
+}
```
adapap, line 130: Can't we just calculate priority based on the context and then sort the turns in order of priority?

battle.go
```diff
@@ -1,115 +1,212 @@
...
+type activePokemon struct {
+	AgentIdx   int
+	PokemonIdx int
 }
 
-// Starts a battle if all pre-conditions are satisfied
-func (b *Battle) Start() {
-	if len(b.Players) < 2 {
-		log.Panicf("not enough players to start a battle")
+// Used to provide data to Agents such that the Agent can't maliciously modify the state of the battle.
```
adapap, line 137: This is a good implementation of the battle context. I guess I haven't worked with OOP enough to understand why the "agent" is any more malicious than this interface considering you can choose what fields to export.

battle.go
```diff
@@ -1,115 +1,212 @@
...
+	b := c.Context
+	p := (*b.Parties[b.agentParties[ap.AgentIdx]]).Pokemon[ap.PokemonIdx]
+	return *p
+}
+func (c *battleContext) SetSelf(idx int) {
+	c.self = idx
+}
+func (c *battleContext) Self() int {
+	return c.self
+}
```
adapap, line 174: Not sure what this is.

battle.go
```diff
@@ -1,115 +1,212 @@
...
+		if !contains(team, ap.AgentIdx) {
+			opponents = append(opponents, i)
 		}
 	}
-	// Cleanup after battle is over / ask to start new battle
+
+	return opponents
+}
+
+// An abstration over all possible actions an `Agent` can make in one round. Each Pokemon gets one turn.
```
adapap, line 200: Continuing on my perspective from earlier, Turn should be a part of the Agent rather than on the Pokemon.

adapap (commented):
It's an interesting approach, but aside from reassigning the functionality to different objects, I do not think this simplifies the end user implementation.

battle.go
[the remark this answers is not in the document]
dyc3: Active pokemon are ones that are currently on the battlefield. I'll update the documentation to reflect that.

battle.go
```diff
@@ -1,115 +1,212 @@
...
+	b.State = BATTLE_IN_PROGRESS
+	for _, p := range b.Parties {
+		p.SetActive(0)
 	}
-	b.Players = append(b.Players, a...)
+	return nil
 }
 
-// Removes a player by their ID from the battle
-func (b *Battle) RemoveAgent(agentID int64) {
```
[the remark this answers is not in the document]
dyc3: That logic should go in the actual implementation of the Agent. We don't have to care about that, because `Act` can block for however long it wants.

battle.go
```diff
@@ -1,115 +1,212 @@
...
 const (
 	BEFORE_START BattleState = iota
-	BATTLE_START
-	BEFORE_PLAYER1_TURN
-	AFTER_PLAYER1_TURN
-	BEFORE_PLAYER2_TURN
-	AFTER_PLAYER2_TURN
+	BATTLE_IN_PROGRESS
 	BATTLE_END
 )
```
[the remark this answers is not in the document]
dyc3: In a round of a Pokemon battle, turns for each pokemon are queued up, and then the game decides what order they execute in. There is no need for the agent to know when turns are happening, because the agent is only responsible for making decisions based on input.

battle.go
dyc3: That is precisely what should happen here. I left it out intentionally for somebody else to implement.

`turns` maps active pokemon to their planned actions. We can't sort the turns directly because they depend on the order of the active pokemon.

battle.go
```diff
@@ -1,115 +1,212 @@
...
-func (b *Battle) AddAgents(a ...*Agent) {
-	if len(b.Players)+len(a) > 2 {
-		log.Panicf("battle cannot have more than two agents connected")
+func (b *Battle) Start() error {
+	// TODO: validate the battle, return error if invalid
+
+	// Initiate the battle! Send out the first pokemon in the parties.
+	b.State = BATTLE_IN_PROGRESS
+	for _, p := range b.Parties {
+		p.SetActive(0)
```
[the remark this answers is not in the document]
dyc3: If user input is required, `Act` is where that will be received. We don't handle that logic.

battle.go
dyc3: Updated documentation

battle.go
```diff
@@ -1,115 +1,212 @@
...
-	ShiftSet bool // shift or set battle style for NPC trainer battles
-	State    BattleState
-	Players  []*Agent
+	Weather      int  // one of the 6 in-battle weather conditions
+	ShiftSet     bool // shift or set battle style for NPC trainer battles
+	State        BattleState
+	Parties      []*Party
+	Agents       []*Agent
+	agentParties map[int]int // Maps agent indexes to party indexes
+	teams        [][]int     // An array of teams, which are arrays of Agents. Used to derive allies and opponents
```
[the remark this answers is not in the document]
dyc3: The index of an Agent is it's ID.

The primary reason for agents to not manage their parties is so that they can't cheat by directly modifying the values of pokemon. That's why `battleContext` provides _copies_ of the data to the agent, and not pointers.

In a double battle, 2v1 trainers is a common scenario, where the lone trainer has a party, and the opponents have 2 different parties. This implementation covers this, others like it pretty nicely. It also allows us to expand pretty easily to doing triple battles, or battle royales, if we want to expand the scope to later pokemon generations.

PR #28, opened 2021-02-09
add link to documentation in README

dyc3:


carpenterd777 (approved):
cool, does this automatically update from the source code?

dyc3:
It should, but i don't know what gets them to invalidate their cache

PR #41, opened 2021-02-12
Implement RunTurn

dcarpenter31:


battle.go
```diff
@@ -66,6 +68,15 @@ const (
 	BATTLE_END
 )
 
+// Indicates the meaning of attempting to take the 'Run' action during this battle.
+type RunSemantics int
+
+const (
+	DISALLOWED = iota
```
dyc3, line 75:
```suggestion
	DISALLOWED RunSemantics = iota
```

battle.go
```diff
@@ -228,3 +248,13 @@ type FightTurn struct {
...
 }
+
+// Indicates that:
+//      in a multibattle, the Team this Agent belongs to will forfeit.
+//      in a wild Pokemon battle, the Agent will attempt to escape.
+// Battles against NPC Trainers cannot be forfeit or escaped.
+type RunTurn struct{}
+
+func (turn RunTurn) Priority() int {
+	return 7 // outside the legal range for any FightTurn
```
dyc3, line 259: Why 7?

battle.go
```diff
@@ -116,6 +127,8 @@ func (b *Battle) SimulateRound() {
 		switch t := turns[apIdx].(type) {
 		case FightTurn:
 			fmt.Printf("TODO: Implement fight %v\n", t)
+		case RunTurn:
+			fmt.Printf("Still implementing")
```
dyc3, line 131: (nit)
```suggestion
			fmt.Printf("TODO: Still implementing")
```

battle_test.go
```diff
@@ -17,39 +17,40 @@ func (dumbAgent) Act(b BattleInfo) Turn {
...
+type scaredAgent struct{}
+
+func (scaredAgent) Act(b BattleInfo) Turn {
+	return RunTurn{}
 }
 
-func TestBattleOneRound(t *testing.T) {
-	a1 := Agent(dumbAgent{})
-	a2 := Agent(dumbAgent{})
+func battleSetup(a Agent) Battle {
```
dyc3, line 27: (nit) I would prefer this to this to be something like `simpleBattleSetup`

dyc3 (changes requested):

battle.go
carpenterd777: 6 is the maximum priority for a FightTurn, I'll change the comment to be clearer

carpenterd777: actually I'll just go make a constant `MAX_PRIORITY` and return that +1

adapap: I added these constants in #43

carpenterd777: since we discussed that the priority of turn is a separate concept from the priority of a move, I'm thinking that `RunTurn`'s priority should be 3

`RunTurn` > `SwitchTurn` > `ItemTurn` > `FightTurn` is what I'm thinking

carpenterd777:
Since we're still iterating over the battle architecture, and I'm not satisfied with my attempts to implement RunTurn in a way that covers cases where the turn is not passed by selecting Run, I'm going to drop this issue for now and move to something else.

PR #42, opened 2021-02-12
Implement ItemTurn, basic item usage

adapap:
Added some of the groundwork for using items in a Pokemon battle. Items are just as, if not more, diverse than all the moves in Pokemon, so a lot of functionality will need to be implemented to handle all item effects. This PR intends to serve as a base layer for using items without too much implementation of item effects (as a challenge to other contributors). Some base contributions include handling status effects, managing stat changes, or interacting with moves/PP. Only a subset of items can actually be used in a battle, so we need to figure out which categories are actually applicable for this library.


pokemon.go
```diff
@@ -12,13 +12,45 @@ type Pokemon struct {
...
+
+// Constants for status effects on a Pokemon
+const (
+	STATUS_BURN = 1 << iota
+	STATUS_FREEZE
+	STATUS_PARALYZE
+	STATUS_POISON
+	STATUS_BADLY_POISON
+	STATUS_SLEEP
+)
```
dyc3, line 42: These were added in the other PR, so these should be removed

pokemon.go
```diff
@@ -51,3 +86,46 @@ func (p *Pokemon) String() string {
...
+	}
+}
+
+// Cures a status ailment from a Pokemon.
+func (p *Pokemon) CureStatusEffect(status uint) {
+	p.StatusEffects &= ^status
+}
+
+// Modifies one of the six base stats for a Pokemon
+func (p *Pokemon) ModifyStat(stat, stages int) {
```
dyc3, line 105: (nit)
```suggestion
func (p *Pokemon) BoostStat(stat, stages int) {
```

dyc3 (changes requested):
This contains too many changes to the code that are irrelevant to what the PR is trying to do. In particular, most of the changes to the codegen script. It makes it really hard to actually review the PR for what it's trying to implement: ItemTurn. Those should go in a separate PR, and large refactorings should at least have an explanation in the PR.

Don't we need to keep these so we can reference them?
- `item_fling_effects.csv`
- `item_game_indices.csv`
- `item_categories.csv`
- `item_pockets.csv`

battle.go
```diff
@@ -228,3 +238,14 @@ type FightTurn struct {
 func (turn FightTurn) Priority() int {
 	return 0
 }
+
+// An item turn has the a higher priority than any move.
+type ItemTurn struct {
+	item   *Item    // Which item is being consumed
+	target *Pokemon // The target Pokemon for the item
```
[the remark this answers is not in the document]
adapap: Why? We can validate that the Pokemon is part of active Pokemon when consuming the item. Keeping track of indices vs. direct references to Pokemon is better for UX imo?

data/gen.go
```diff
@@ -225,60 +228,40 @@ func main() {
...
+			// fmt.Sprintf("CurrentHP:%d,\n", uint(0)) +
+			// fmt.Sprintf("HeldItem:%v,\n", "new(Item)") +
+			fmt.Sprintf("Moves:%s,\n", "[4]*Move{}") +
+			// fmt.Sprintf("Friendship:%d,\n", uint8(0)) +
+			// fmt.Sprintf("OriginalTrainerID: %d,\n", uint16(0)) +
+			"},\n"
 	}
-
-	_, err = output.WriteString("}\n\n" +
+	output += "}\n\n" +
```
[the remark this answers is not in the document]
adapap: Disagree:
- String concatenation is not a bad idea but in this case also is not necessary, as you can rewrite each line to `output += text`
- The only time this file is read is when using `go generate` so neither memory nor performance is an issue

The rationale for changing to string instead of writing to file is to avoid the missing error checks that linters are expecting for each `output.WriteString(...)`. An alternative solution is a helper function such as `writeOutput(s)`.

pokemon.go
adapap: Is boost ambigious because this can take negative stages for a debuff?

adapap:
> This contains too many changes to the code that are irrelevant to what the PR is trying to do. In particular, most of the changes to the codegen script. It makes it really hard to actually review the PR for what it's trying to implement: ItemTurn. Those should go in a separate PR, and large refactorings should at least have an explanation in the PR.
> 
> Don't we need to keep these so we can reference them?
> 
> * `item_fling_effects.csv`
> * `item_game_indices.csv`
> * `item_categories.csv`
> * `item_pockets.csv`

I could move the generate code into a separate PR but I just wanted to have it be cleaner as I work with it. Sure, some of the changes are subjective but it fixes missing error checks and improves readability for something not performance-critical.

Addressing the deleted files:
- `item_fling_effects.csv`: This just maps the index to the fling effect name. I added these as constants (e.g. `FLING_BURN`).
- `item_game_indices.csv`: This maps the global item ID from `items.csv` to the internal game ID for each generation. I found this unnecessary as we don't care about the item ID in any specific game.
- `item_categories.csv`: Similar to fling effects, just tracks identifiers that are better represented as constants.
- `item_pockets.csv`: Is there any reason why we need to know an item's pocket? Not relevant to battles at all.

dyc3 (changes requested):
Please move the codegen changes to a separate PR. Also, tests are failing.

battle.go
```diff
@@ -228,3 +238,14 @@ type FightTurn struct {
...
+	move   *Move    // Reference to a move for certain items
+}
+
+func (turn ItemTurn) Priority() int {
+	return PRIORITY_MAX
```
dyc3, line 250: Priority should be 1
```suggestion
	return 1
```

battle.go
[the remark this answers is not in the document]
dyc3: Because the agent _should_ receive a copy of the battle data, so the agent can't possibly know the pointer to the real pokemon. If it _can_ do that, then it's a bug.

dyc3 (changes requested):

pokemon.go
dyc3: Oh, you're right, maybe it is. I think it should be something other than modify, because it makes me think it modifies the underlying stat, not adding buffs/debuffs

adapap: https://bulbapedia.bulbagarden.net/wiki/Stat#Stat_modifiers Seems like modify is the word.

battle.go
```diff
@@ -214,6 +221,9 @@ func (c *battleContext) Opponents() []int {
 }
 
 // An abstration over all possible actions an `Agent` can make in one round. Each Pokemon gets one turn.
+const PRIORITY_MAX = 5
+const PRIORITY_MIN = -7
+
```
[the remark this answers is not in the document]
dyc3: These lines are still present

dyc3 (changes requested):

data/gen.go
```diff
@@ -100,6 +100,7 @@ func createCodeOutput(s string) {
 	}
 	_, err = file.WriteString("// Code generated - DO NOT EDIT.\n" +
 		"// Regenerate with `go generate`.\n\n" +
+		"// go:generate go run data/gen.go\n" +
```
dyc3, line 103: This should remain in pokedex.go, because `go generate` should work without the presence of the generated code.

You gotta start putting these things in separate PRs. It makes it really hard to review your PRs, and a few pretty significant bugs have gotten through as a result. Making refactors when implementing a feature usually means that you don't have what is required to implement the feature, and chances are somebody else needs it too. Make the refactor in a different PR, stating your reason, and then rebase your feature's branch on main when it gets merged in.

battle.go
```diff
@@ -131,16 +133,29 @@ func (b *Battle) SimulateRound() []Transaction {
...
 	for _, transaction := range transactions {
 		switch t := transaction.(type) {
 		case DamageTransaction:
 			(*t.Target).CurrentHP -= t.Damage
+		case ItemTransaction:
+			if t.Item.Category == ItemPPRecovery {
+				t.Item.UseMoveItem(t.Target, t.Move)
+			} else {
+				t.Item.UseItem(t.Target)
+			}
```
dyc3, line 158: I'd prefer if the logic for processing the transaction was put here instead of using methods that are somewhere else. It would be better if public facing methods created a transaction that the battle processes. That way we have only one possible way to modify the state. And if we play back all the transactions for a battle in order, the resulting battle state is guaranteed to be the same every time, even if the user wants to cheat (outside the context of an `Agent` ofc).

Also, this feels backwards, like it should be `t.Target.UseItem(t.Item)` or `t.Move.UseItem(t.Item)`. This would also make more sense for when a pokemon auto-eats a berry

battle.go
```diff
@@ -232,3 +259,13 @@ func (t DamageTransaction) BattleLog() string {
...
 }
+
+type ItemTransaction struct {
+	Target *Pokemon
+	Item   *Item
+	Move   *Move
+}
+
+func (t ItemTransaction) BattleLog() string {
+	return fmt.Sprintf("%s used on %s.", t.Item.Name, t.Target.GetName())
```
dyc3, line 270: Probably not necessary to change now, but this doesn't really provide a whole lot of information. Doesn't describe stuff like how much hp was restored, or if status effects were removed.

dyc3 (changes requested):

battle.go
adapap: There's a lot of items with a ton of different effects. If they don't get delegated to different methods, this file will become excessively long as its purpose is not to handle dealing with items.

I don't understand what you mean by "public facing methods created a transaction". Do you mean something like: `pokemon.UseItem(item *Item) Transaction`?

I do agree with the naming order, I will change that.

battle.go
adapap: Not sure how this should be handled. In theory, using an item and its effect are two separate logs. Using a potion and a healing move both produce an effect that heals a Pokemon, which leads me to believe it should be a separate transaction like `HealTransaction` or similar.

battle_test.go
```diff
@@ -94,6 +108,38 @@ func TestBattleOneRound(t *testing.T) {
...
+		},
+	}
+	for _, tt := range logtest {
+		got := tt.turn.BattleLog()
+		if got != tt.want {
+			t.Errorf("Expected battle log to be %s, got %s", tt.want, got)
+		}
+	}
+	if b.parties[0].pokemon[0].CurrentHP != 30 {
+		t.Errorf("expected potion to be used on first Pokemon")
```
dyc3, line 139: should be "expected pokemon's hp to be ..., got ..."

dyc3 (changes requested):

dyc3 (changes requested):
Test coverage is way too low.

This PR tries to implement the usage of multiple types of items, and also held items affects damage. Instead of trying to increase coverage by writing tests (because that would require _more_ code, and this PR needs to get merged soon), remove code and focus on a single item's implementation. Leave the others for another PR.

My suggestion:
- focus on just potions
- remove held item code
- remove PP restore
- remove status cures
- remove stat modifications

If this PR's patch coverage gets >97%, it's an automatic approval from me, the **absolute minimum is 90%**.

dyc3 (approved):
It looks like your changes in pokemon.go is killing your coverage (which isn't covered anyway) so it's good. I'll merge it in.

PR #50, opened 2021-02-12
sort turns

### LINKED ISSUE #31: Sort turns by priority
In a Pokemon battle, certain types of turns are supposed to occur before others.

Relevant code: https://github.com/StevensSEC/pokemonbattlelib/blob/07c4cf0a2732d5aafef2b9bc722b7e4e08804ae6/battle.go#L118

Sort `turnOrder` in descending order by `Priority()` on the corresponding turns.

dyc3:
- refactor code into sortTurns
- Implement minimal turn order sorting

closes #31

### DIFF
--- a/battle.go
+++ b/battle.go
@@ -106,11 +106,7 @@ func (b *Battle) SimulateRound() {
 		turns[i] = turn
 	}
 
-	turnOrder := []int{}
-	// TODO: determine the correct order based on priority and other factors
-	for i := range active {
-		turnOrder = append(turnOrder, i)
-	}
+	turnOrder := sortTurns(active, turns)
 
 	for _, apIdx := range turnOrder {
 		switch t := turns[apIdx].(type) {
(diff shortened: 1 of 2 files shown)


battle.go
```diff
@@ -143,6 +141,67 @@ func (b *Battle) derefActivePokemon(ap activePokemon) *Pokemon {
 	return (*b.Parties[b.agentParties[ap.AgentIdx]]).Pokemon[ap.PokemonIdx]
 }
 
+// A type that is necessary in order to implement the `Interface` interface, which is used by the sort package to sort.
+type apTurnOrder struct {
+	battle *Battle
+	active []activePokemon
+	turns  map[int]Turn
```
adapap, line 148: We can use [slice sorting](https://yourbasic.org/golang/how-to-sort-in-go/#sort-with-custom-comparator) to avoid creating a new struct here, since we are sorting a slice of structs.

battle.go
```diff
@@ -143,6 +141,67 @@ func (b *Battle) derefActivePokemon(ap activePokemon) *Pokemon {
...
+	tA, tB := t.turns[a], t.turns[b]
+	if reflect.TypeOf(tA) == reflect.TypeOf(tB) {
+		switch tA.(type) {
+		case FightTurn:
+			// speedy pokemon should go first
+			return pA.Stats[5] > pB.Stats[5]
+		}
+	} else {
+		// make higher priority turns go first
+		return tA.Priority() > tB.Priority()
```
adapap, line 192: This is probably fine for now, but items like the [Quick Claw](https://bulbapedia.bulbagarden.net/wiki/Quick_Claw) will probably cause some additional checks here. My suggestion as mentioned in the meeting is to keep track of some additional flags on Pokemon/in the battle, with quick claw being one of them.

adapap (approved):
Can simplify the sorting functionality, but otherwise LGTM

battle.go
dyc3: Not with the current way the data is laid out. The order of active pokemon is relied upon.

battle.go
dyc3: Yes, but that check would not go on this line. The priority check here indicates which _type_ of turns should go first. For example, IIRC Item turns should always go before Fight turns.

PR #52, opened 2021-02-12
refactor out the necessity of LinkAgentParty, fixes #44

### LINKED ISSUE #44: refactor out LinkAgentParty
We decided in a GBM that LinkAgentParty was not necessary, and just increases complexity.

dyc3:


battle.go
```diff
@@ -6,19 +6,18 @@ import (
...
-	State        BattleState
-	Parties      []*Party
-	Agents       []*Agent
-	agentParties map[int]int // Maps agent indexes to party indexes
-	teams        [][]int     // An array of teams, which are arrays of Agents. Used to derive allies and opponents
+	Weather  int  // one of the 6 in-battle weather conditions
+	ShiftSet bool // shift or set battle style for NPC trainer battles
+	State    BattleState
+	Parties  []*Party
+	teams    [][]int // An array of teams, which are arrays of Party ids. Used to derive allies and opponents
```
adapap, line 13: (nit) Can team also just be a field on Party? If it's tracking the Agent presumably it can also track the team that the Agent is on?

battle.go
```diff
@@ -165,10 +146,10 @@ type battleContext struct {
 }
 
 func (c *battleContext) getTeam() []int {
-	selfAgent := c.ActivePokemon[c.self].AgentIdx
+	selfParty := c.ActivePokemon[c.self].PartyIdx
 	var team []int
 	for _, t := range c.Context.teams {
```
adapap, line 151: This might be cleaner if we move team to a field on Party.

carpenterd777 (approved):

adapap (approved):
LGTM - teams is also a bit of index management now that I look at it

battle.go
dyc3: You're probably right, but lets leave it out of this PR.

PR #56, opened 2021-02-13
add GetMove, add moves to pokemon in test cases

dyc3:


battle_test.go
```diff
@@ -42,12 +42,15 @@ func TestBattleOneRound(t *testing.T) {
 	party1 := Party{
 		Agent: &a1,
 	}
+	pound := GetMove(1)
 	pkmn1 := GetPokemon(4)
+	pkmn1.Moves[0] = &pound
```
adapap, line 47: Can we make this a `pokemon.SetMove(i int, move *Move)`?

adapap (changes requested):

carpenterd777 (approved):

battle_test.go
dyc3: Yeah, I agree, but it should go in another PR imo

Issue #34, opened 2021-02-09
Validate teams at beginning of battle


dyc3:
Relevant code: https://github.com/StevensSEC/pokemonbattlelib/blob/07c4cf0a2732d5aafef2b9bc722b7e4e08804ae6/battle.go#L53-L57

adapap:
Responding to #57, teams should be validated in `battle.Start`. The check should be that there exists (at the moment) exactly two teams with 1 or more agents on each team.

https://github.com/StevensSEC/pokemonbattlelib/blob/5d0270a39720dd9784a43f33ee9c7e1573ab71db/battle.go#L64-L73

dyc3:
Unassigning due to inactivity.

PR #57, opened 2021-02-13
Battle rework v3

### LINKED ISSUE #60: Properties of FightTurn are inaccessible publicly, therefore making it impossible to use the turn outside of the package.
When using the library as intended, it's currently impossible for agents to initialize or modify the values for FightTurn.

adapap:
This PR attempts to change the underlying structure of battles, parties, and turns in order to become as independent on index management and use less technical naming conventions for library functions. Some changes and their arguments:

- Constructors for creating battles, parties, and Pokemon were added. This allows us to add (extensible) default options to the struct without asking the user to add all required fields themselves. It also allows us to make `party` a package-only type to minimize illegal party manipulation.

- `party.go` was created to separate party management from battles, which simplifies testing and makes for better code organization.

- Teams were moved to a field on parties as they can be inferred from the battle to calculate allies/opponents. Users are currently in charge of maintaining unique team values (as they were previously), but now parties are directly assigned a team in their creation.

- `pokemon` and `activePokemon` are two distinct fields now managed by the party. This was an important change to reduce index management, as the index can always be found by checking which pair of pointers are equal. I think in practice this lead to more intuitive development, but this is somewhat subjective.

- `Context` replaced `battleInfo` and `BattleContext`. It contains (de-referenced) copies of the data that the library uses internally, so the end user can just check the stats of a Pokemon without needing to figure out what index is needed. I believe this still maintains immutability of the battle state from the Agent.

- Turns now use a `TurnContext` object to avoid needing to keep track of indices during Pokemon actions and when determining the sorted order using priority, stats, etc.

The underlying idea behind these changes is to have every library function be understood by a casual Pokemon player with limited or no programming experience. Definitely interested in the feedback!

fixes #60


dyc3:
This will invalidate #34

battle.go
```diff
@@ -55,214 +23,128 @@ const (
...
+		case FightTurn:
+			log.Printf("TODO: Implement fight\n")
+		default:
+			log.Panicf("Unknown turn of type %v", t)
 		}
 	}
+}
 
-	return allies
+type Context struct {
```
dyc3, line 116: (nit) to avoid name conflicts with other packages
```suggestion
type BattleContext struct {
```

Does this struct need to be public?

battle.go
```diff
@@ -55,214 +23,128 @@ const (
...
-func (c *battleContext) Opponents() []int {
-	team := c.getTeam()
 
-	opponents := []int{}
-	for i, ap := range c.ActivePokemon {
-		if !contains(team, ap.PartyIdx) {
-			opponents = append(opponents, i)
-		}
+// Gets the current context for a pokemon to act (perform a turn)
+func (b *Battle) GetContext(party *party, pokemon *Pokemon) *Context {
```
dyc3, line 124: Does this need to be public?

pokedex.go
```diff
@@ -1,14 +1,17 @@
 package pokemonbattlelib
 
+import "log"
+
 //go:generate go run data/gen.go
 
-// Get a new pokemon by nation dex number.
-func GetPokemon(natdex uint16) Pokemon {
+// Creates a new Pokemon given its national dex number
+func NewPokemon(natdex uint16) *Pokemon {
```
dyc3, line 8: This either needs to explicitly copy the pokemon, or should not return a pointer, because then the Pokemon being referenced is the one inside of `ALL_POKEMON`

util.go
```diff
@@ -1,10 +1,10 @@
...
-	return false
-}
+// func contains(s []int, e int) bool {
+// 	for _, a := range s {
+// 		if a == e {
+// 			return true
+// 		}
+// 	}
+// 	return false
+// }
```
dyc3, line 10: If this is no longer needed, then you should remove it.

dyc3 (changes requested):

battle.go
```diff
@@ -55,214 +23,128 @@ const (
...
 		}
 	}
+}
 
-	return allies
+type Context struct {
+	Battle    Battle
+	Pokemon   Pokemon
+	Allies    AgentParties
+	Opponents AgentParties
```
dyc3, line 120: Needs to be documented

party.go
```diff
@@ -0,0 +1,84 @@
...
+// Sets a Pokemon to be active by its index in a party (0-5)
+func (p *party) SetActive(i int) {
+	if p.IsActivePokemon(i) {
+		log.Panicf("pokemon is already out on the battlefield")
+	}
+	p.activePokemon = append(p.activePokemon, p.pokemon[i])
+}
+
+// Sets a Pokemon to be inactive by its index in a party (0-5)
+func (p *party) SetInactive(i int) {
```
dyc3, line 47: This function is not tested

dyc3 (changes requested):

dyc3:
Made a playground to check if agents could modify the pokemon directly. Seems to check out, needs to be put in a test.

<details>
<summary>Playground</summary>
```go
package main

import (
	"fmt"

	lib "github.com/StevensSEC/pokemonbattlelib"
)

type dumbAgent struct{}

// Blindly uses the first move on the first opponent pokemon.
func (a dumbAgent) Act(ctx *lib.Context) lib.Turn {
	// You can use `a` (reference to self) for self-targeting turns
	for agent, party := range ctx.Opponents {
		for i := range party {
			return lib.FightTurn{
				Agent:  agent,
				Move:   0,
				Target: i,
			}
		}
	}
	panic("no opponents found")
}

type badBoy struct{}

func (badBoy) Act(ctx *lib.Context) lib.Turn {
	for agent, party := range ctx.Opponents {
		p := party[0]
		p.CurrentHP = 765
		p.Stats[lib.STAT_ATK] = uint(123)
		fmt.Printf("setting %s\n", p.GetName())
		for i := range party {
			return lib.FightTurn{
				Agent:  agent,
				Move:   0,
				Target: i,
			}
		}
	}
	panic("no opponents found")
}

func check(err error) {
	if err != nil {
		fmt.Printf("Error: %v", err)
	}
}

func main() {
	a1 := lib.Agent(badBoy{})
	a2 := lib.Agent(dumbAgent{})
	b := lib.NewBattle()
	party1 := lib.NewParty(&a1, 0)
	pkmn1 := lib.NewPokemon(4)
	party1.AddPokemon(pkmn1)
	party2 := lib.NewParty(&a2, 1)
	pkmn2 := lib.NewPokemon(7)
	party2.AddPokemon(pkmn2)
	b.AddParty(party1, party2)
	err := b.Start()
	check(err)

	b.SimulateRound()

	tpkmn := b.Parties[1].GetActivePokemon()[0]
	fmt.Printf("results: %s\n", tpkmn.GetName())
	fmt.Printf("current hp: %d\n", tpkmn.CurrentHP)
	fmt.Printf("result: %d\n", tpkmn.Stats[lib.STAT_ATK])
}
```
</details>

battle.go
adapap: Yes, this type is a parameter in `agent.Act(ctx)`.

carpenterd777:
what is `c.out` for?

adapap:
> what is `c.out` for?

Generated test coverage file, just removed. See #64.

move.go
```diff
@@ -38,6 +38,16 @@ type Move struct {
 	Accuracy int
 }
 
+// Creates a new Pokemon move given its move ID
+func NewMove(id int) Move {
```
dyc3, line 42: (nit) Implies that the user is creating a brand new move, one that is not in the pokedex data.

`GetMove` already exists in pokedex. pokedex is currently being treated as a database kinda, where the functions inside are specifically ones to retrieve information from that database. It'll be easier to make it a sub-package later, if we want to go that route.

party.go
```diff
@@ -0,0 +1,84 @@
...
+	"log"
+)
+
+// A Pokemon party. Can hold up to 6 Pokemon. Also manages how many pokemon are out on the battlefield.
+type party struct {
+	Agent         *Agent     // The agent that has control over this party
+	pokemon       []*Pokemon // The Pokemon in the party
+	activePokemon []*Pokemon // The pokemon in the party that are out on the battlefield
+	team          int        // The team that this party belongs to
+}
```
dyc3, line 14: Suppose a user wanted to do multiple battles in a row with the same party of pokemon. Would the user have to use their own type to keep track of the party, and then convert it into this one when they need to put it into a new battle?

In that scenario, we can't guarantee that `team` will remain valid if this party is directly reused. Nor can we guarantee that `Agent` doesn't have state that would make it behave differently than if it was freshly created.

pokemon.go
```diff
@@ -39,6 +41,17 @@ const (
 	STATUS_SLEEP
 )
 
+// Creates a new Pokemon given its national dex number
+func NewPokemon(natdex uint16) Pokemon {
```
dyc3, line 45: (nit) same issue as `NewMove`

pokemon_test.go
```diff
@@ -2,12 +2,13 @@ package pokemonbattlelib
...
+	p := NewPokemon(8)
+	if p.NatDex != 8 {
+		t.Errorf("expected Pokemon national dex to be 8, received %v", p.NatDex)
 	}
-	if p.GetName() != "Bulbasaur" {
-		t.Fail()
+	if p.GetName() != "Wartortle" {
+		t.Errorf("expected Pokemon name to be Wartortle, received %v", p.GetName())
 	}
 }
```
dyc3, line 13: This test replaces another test when it should not. It also tests 2 different things.

battle_test.go
```diff
@@ -6,63 +6,65 @@ import (
...
+		got := party.GetActivePokemon()
+		if !reflect.DeepEqual(got, map[int]Pokemon{0: *party.pokemon[0]}) {
 			t.Fatalf("Must send out first pokemon in each at the beginning of the battle. Party %d gave: %v", p, got)
 		}
 	}
 	b.SimulateRound()
+	party1.SetInactive(0)
+	if len(b.parties[0].activePokemon) != 0 {
+		t.Error("expected party 1 to have no active Pokemon")
+	}
```
dyc3, line 67: IMO, this should go in a separate test just to test `SetActive` and `SetInactive`. It also conflicts with the fight turn PR I'm working on.

dyc3 (changes requested):
We're definitely getting somewhere.

The biggest problem I've had with this PR was when I was making the playground I posted earlier. It was not possible to observe the state of the battle by inspecting the public properties of `Battle`.

My 2nd biggest problem is that it increases the complexity of `dumbAgent`. I don't like how you can't know the keys/values of the `Allies` and `Opponents` map without iterating over it. In the current version, the Agents are told identifiers for pokemon, and then they can access them with a helper function `GetPokemon` on `BattleInfo`.

move.go
carpenterd777: I interpreted it to be like the way the `new` keyword is used in say Java, to imply that's a constructor. Can we make functions in go that are the same as the name of a type?

adapap: From what I've seen in Go, `NewMyType` functions are used to create initial values for custom types. For some examples, see Go's [image](https://golang.org/pkg/image/) package and [this blurb](https://golang.org/doc/effective_go.html#allocation_new) about creating new types. If `move` became an unexported type, then we can use `Move()` as the constructor.

I think the generated data should serve as a database and not as a base struct value. The idea with having constructors is that we can pass in parameters (e.g. custom PP, IVs for Pokemon, metadata for battles) and not have users concerned as much about the internal representation/implementation of `Move`/`Battle`. We just have to enforce standard (#61 likely relevant).

party.go
adapap: Since the party holds direct references to the Pokemon I was concerned about immutability during battle state so I wanted to keep it unexported. There's probably better solutions here, but we potentially can have a party reset helper similar to a PokeCenter (items are tricky in that case). Depending on #61 it might become trivial to loop through battles from some initial setup.

I still think teams should be managed by the user, but each battle/round asserts that there is a valid team configuration.

adapap:
Pretty happy with the organization of this now. When each agent is called to act, they are given an array of `target` objects which contain `Pokemon` and `Team` exported, but internally also have `party` and `partySlot` so that we can correctly reference the target. This allows the target to simply be passed into a turn as such:

`FightTurn{Move: 0, Target: target}`

Team is included in the target which allows for quickly determining allies/opponents since agents are expected to know their team and party. Since this doesn't make it too easy to determine if a Pokemon belongs to the acting agent's party, we can make targets not include the agent's party and have a `SelfTarget` array of Pokemon belonging to the agent.

The main benefit(s) of this are that the user needs no index management for parties. Aside: I think something similar can be adapted for moves to not rely on its index.

Included in the latest change is the `playground.go` file which shows that agents cannot modify Pokemon, but can read their stats and other data. Here is how the player can query the first Pokemon: `b.GetTargets()[0].Pokemon`

dyc3:
> Included in the latest change is the playground.go

The playground is not intended to be committed. It's intended to be easily sharable, but not a replacement for tests. I guess I should have documented that

carpenterd777 (approved):

move.go
dyc3: > I think the generated data should serve as a database and not as a base struct value.

Yes, precisely. This function is being used to grab a move from the database, and use it as a template.
The name `NewMove` should be reserved for creating a move that is not in the database (which is not yet implemented).

dyc3 (changes requested):
`playground.go` should not be committed.

dyc3 (changes requested):
It's currently not possible for an agent to determine if a target is an ally or opponent. I'd like to see `BattleContext` implement helper functions for this (like `BattleInfo` currently does).

adapap:
> It's currently not possible for an agent to determine if a target is an ally or opponent. I'd like to see `BattleContext` implement helper functions for this (like `BattleInfo` currently does).

Targets contain a `Team` field. Agents should know their team, so this can be inferred by checking if the team matches. What kind of helper are you looking for?

dyc3:
But agents _don't_ know what team they are on. It's not provided in `BattleContext`

The helper functions I'm looking for are `Allies()` and `Opponents()`, like the ones `BattleInfo` provides in battles v2.

adapap:
@dyc3 Added more info to the battlecontext object. Technically, teams/targets and allies/opponents can both work separately so there is some redundancy here. Let me know if you want to simplify to one of the other.

battle_test.go
```diff
@@ -80,30 +75,117 @@ func TestBattleOneRound(t *testing.T) {
...
+	a2 := Agent(dumbAgent{})
+	party1 := NewParty(&a1, 0)
+	pkmn1 := GetPokemon(4)
+	party1.AddPokemon(&pkmn1)
+	party2 := NewParty(&a2, 1)
+	pkmn2 := GetPokemon(7)
+	pkmn3 := GetPokemon(9)
+	party2.AddPokemon(&pkmn2, &pkmn3)
+	b := NewBattle()
+	b.AddParty(party1, party2)
```
dyc3, line 90: Is it necessary to initialize a whole battle to test this?

battle_test.go
```diff
@@ -80,30 +75,117 @@ func TestBattleOneRound(t *testing.T) {
...
+	party1 := NewParty(&a1, 0)
+	pkmn1 := GetPokemon(4)
+	pkmn2 := GetPokemon(7)
+	pkmn3 := GetPokemon(11)
+	party1.AddPokemon(&pkmn1, &pkmn2, &pkmn3)
+	b := NewBattle()
+	b.AddParty(party1)
+	p := b.getPokemon(0, 1)
+	if p.NatDex != 7 {
+		t.Errorf("expected Pokemon with dex number 9, received %v\n", p.NatDex)
```
dyc3, line 116:
```suggestion
		t.Errorf("expected Pokemon with dex number 7, received %v\n", p.NatDex)
```

dyc3 (commented):
I've come up with this playground to try out the new API. I've made a new agent that uses a little bit of logic to determine what the best move to use is. I think the Agent API is pretty easy to use. The only part I don't like is the necessity to check if a move is nil, but it's a small issue.

<details>

<summary>Playground</summary>

```go
package main

import (
	"fmt"

	lib "github.com/StevensSEC/pokemonbattlelib"
)

type dumbAgent struct{}

// Blindly uses the first move on the first opponent pokemon.
func (a dumbAgent) Act(ctx *lib.BattleContext) lib.Turn {
	// You can use a (reference to self) for self-targeting turns
	return lib.FightTurn{
		Move:   0,
		Target: ctx.Opponents[0],
	}
}

func check(err error) {
	if err != nil {
		fmt.Printf("Error: %v", err)
	}
}

type smartAgent struct{}

func (smartAgent) Act(ctx *lib.BattleContext) lib.Turn {
	var strongestMove int
	power := 0
	for i, move := range ctx.Pokemon.Moves {
		if move == nil {
			continue
		}
		if move.Power > power {
			power = move.Power
			strongestMove = i
		}
	}
	best := 0
	for i, target := range ctx.Opponents {
		if target.Pokemon.Stats[lib.STAT_DEF] < ctx.Opponents[best].Pokemon.Stats[lib.STAT_DEF] {
			best = i
		}
	}
	return lib.FightTurn{
		Move:   strongestMove,
		Target: ctx.Opponents[best],
	}
}

func main() {
	a1 := lib.Agent(smartAgent{})
	a2 := lib.Agent(dumbAgent{})
	b := lib.NewBattle()
	party1 := lib.NewParty(&a1, 0)
	pkmn1 := lib.GetPokemon(4)
	pkmn1.Stats = [6]uint{30, 10, 10, 10, 10, 10}
	pkmn1.CurrentHP = 30
	pound := lib.GetMove(1)
	chop := lib.GetMove(2)
	pkmn1.Moves[0] = &pound
	pkmn1.Moves[1] = &chop
	party1.AddPokemon(&pkmn1)
	party2 := lib.NewParty(&a2, 1)
	pkmn2 := lib.GetPokemon(7)
	pkmn2.Stats = [6]uint{30, 10, 10, 10, 10, 10}
	pkmn2.CurrentHP = 30
	wingattack := lib.GetMove(17)
	pkmn2.Moves[0] = &wingattack
	party2.AddPokemon(&pkmn2)
	b.AddParty(party1, party2)
	err := b.Start()
	check(err)

	for i := 0; i < 3; i++ {
		fmt.Printf("Turn %d\n", i)
		b.SimulateRound()

		for _, t := range b.GetTargets() {
			fmt.Printf("pokemon: %s\n", t.Pokemon.GetName())
			fmt.Printf("current hp: %d\n", t.Pokemon.CurrentHP)
		}
	}
}
```
</details>

adapap:
> I've come up with this playground to try out the new API. I've made a new agent that uses a little bit of logic to determine what the best move to use is. I think the Agent API is pretty easy to use. The only part I don't like is the necessity to check if a move is nil, but it's a small issue.

We can probably make moves a slice and have a MAX_MOVES constant to stay consistent with parties. That way custom gamemodes with more than 4 moves are possible. This is outside the scope of this PR however.

dyc3 (approved):

PR #58, opened 2021-02-13
Added getters and setters for stats

carpenterd777:
I figured this would be a clearer way to tell whats going on in tests and the like when working with stats.

### DIFF
--- a/battle.go
+++ b/battle.go
@@ -166,7 +166,7 @@ func (t *apTurnOrder) Less(a, b int) bool {
 		switch tA.(type) {
 		case FightTurn:
 			// speedy pokemon should go first
-			return pA.Stats[5] > pB.Stats[5]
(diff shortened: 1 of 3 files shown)


dyc3:
I don't think this is necessary anymore because we have the `STAT_` constants. 

```go
torchic.Stat[STAT_HP]
```

The relevant code that doesn't use these constants needs to be changed.

dyc3:
Like I said before, I don't think this change is necessary. Paging @adapap for opinion

adapap:
Agree that it's not needed because the only time the end user changes stats is when creating a new Pokemon. The only thing I can think of is `p.SetStat(stat, amount)` and `p.GetStat(stat)` but that's probably redundant especially since we would need it for IVs/EVs for consistency.

PR #62, opened 2021-02-14
Table-ize some tests

dyc3:
- fix stringer implementation for Pokemon
- refactor tests to make it easier to add more

I also took the liberty to change `Gender`'s Stringer implementation to just use the characters instead of escaping them.

### DIFF
--- a/gender.go
+++ b/gender.go
@@ -14,9 +14,9 @@ func (g Gender) String() string {
 	if g == Genderless {
 		return ""
 	} else if g == Female {
-		return "\u2640"
+		return "♀"
 	} else if g == Male {
-		return "\u2642"
+		return "♂"
 	} else {
 		panic("Stringing gender reached unhandled condition")
 	}
--- a/pokemon.go
+++ b/pokemon.go
@@ -67,7 +67,7 @@ func (p *Pokemon) HasValidEVs() bool {
 // implement Stringer
 
 // display a Pokemon close to how it would appear in a Pokemon battle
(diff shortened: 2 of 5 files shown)


pokemon.go
```diff
@@ -67,7 +67,7 @@ func (p *Pokemon) HasValidEVs() bool {
 // implement Stringer
 
 // display a Pokemon close to how it would appear in a Pokemon battle
-func (p *Pokemon) String() string {
+func (p Pokemon) String() string {
 	return fmt.Sprintf("%v%v\tLv%d\nHP: %d/%d\n", p.GetName(),
 		p.Gender, p.Level, p.CurrentHP, p.Stats[0])
```
carpenterd777, line 72:
```suggestion
		p.Gender, p.Level, p.CurrentHP, p.Stats[STAT_HP])
```

while we're editing this might as well right?

gender.go
```diff
@@ -14,9 +14,9 @@ func (g Gender) String() string {
 	if g == Genderless {
 		return ""
 	} else if g == Female {
-		return "\u2640"
+		return "♀"
```
carpenterd777, line 17: how did you do that?

adapap: Copy paste the character from the internet 😮

carpenterd777: oh lmao

pokemon.go
dyc3: agree

carpenterd777 (approved):

PR #68, opened 2021-02-15
Battle Log - Solution A

### LINKED ISSUE #49: Tests that check std output pass when they shouldn't
## Current Behavior
Some of the current tests that check the std output are not actually checking it.

**To Reproduce**
If you take the `TestBattleOneRound` and change the expected output, the test will still pass.

**Evironment:**
 - OS: Ubuntu 20.04

**Additional context**

### LINKED ISSUE #51: Battle Log/History
We need some way to let users know what happened in each round of the battle.

Possible solution:
- Have each turn implement a `BattleLog()` method that outputs a string that would be end-user facing (eg. "Bulbasaur used X for -Y HP. It was super effective!"), and also have Turns make data about the result publicly accessible so that users could do animations and stuff. The battle could then just record the turns.

**Additional context**
Also may be necessary to fix #49 .

dyc3:
- adjust Turn priority documentation
- Add result field to Turns to be able to observe changes to battle state

closes #51
indirectly fixes #49
closes #69 

This is one of 2 possible implementations that I created for logging changes in the battle. (See #69)

This version adds a `BattleLog()` function to all Turns, which presents a human readable version of what 
happened in that turn. It also requires that all new turns include a `Result` field with all data that is 
required for a user to be able to observe changes to the battle state. (For example, animating moves).

### Benefits

- Very simple implementation

### Problems

- When we implement stuff like status effects we would have to create an extra turn to describe that change, 
because poison/burn damage happen at the end of the round.
- It'll be slightly harder for a user to process a turn that has multiple changes to the battle state, 
like a `FightTurn` that deals damage and also burns the target.

### DIFF
--- a/battle.go
+++ b/battle.go
@@ -1,6 +1,7 @@
 package pokemonbattlelib
 
 import (
+	"fmt"
 	"reflect"
 	"sort"
 )
@@ -71,7 +72,7 @@ func (b *Battle) Start() error {
 }
 
 // Simulates a single round of the battle.
-func (b *Battle) SimulateRound() {
+func (b *Battle) SimulateRound() []Turn {
 	active := b.getActivePokemon()
 
 	context := battleContext{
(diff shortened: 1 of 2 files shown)


adapap (commented):
Prefer #69 because `Turn` should not be confused with a change in the battle state, and because some events are incompatible (weather change, failure to use item, etc)

battle.go
```diff
@@ -259,16 +267,32 @@ func (c *battleContext) Opponents() []int {
...
+		Damage uint
+	}
 }
 
 func (turn FightTurn) Priority() int {
 	return 0
 }
+
+func (turn FightTurn) BattleLog() string {
+	return fmt.Sprintf("%s used %s on %s for %d damage.",
```
carpenterd777, line 292: I think this a good start, but I think it overgeneralizes what the result of a FightTurn might be. What about moves that heal? What about moves that change stats? How will this be handled in `BattleLog()` in the future?

dyc3:
Closing this in favor of #69

Issue #12, opened 2021-01-26
Example Pokemon Battles


dyc3:
We need test cases to ensure the accuracy of our Pokemon battle implementation.

Eventually, we will turn these into automated test cases.

### Tasks

- [ ] Describe 3 Pokemon battles from start to finish, clearly documenting the results for each turn with the reason for why it happened, and put it in the documentation (`docs/example-battles.md`)
    - If possible, include the RNG seed at the start of the battle, damage numbers, buffs, debuffs, etc. The more information the better.
    - [ ] One battle should be as simple as possible, avoiding edge cases, between 2 low level Pokemon (one on each side).
    - [ ] One battle should include super-effective/not-very-effective type matchups.
    - [ ] The complexity of the last battle is up to your discretion.

ronduma:
[low-level-battle.txt](https://github.com/StevensSEC/pokemonbattlelib/files/5985593/low-level-battle.txt)

dyc3:
Unassigning due to inactivity.

dyc3:
This is no longer relevant to the current state of the project.

PR #69, opened 2021-02-15
Battle Log - Solution B

### LINKED ISSUE #49: Tests that check std output pass when they shouldn't

### LINKED ISSUE #51: Battle Log/History

dyc3:
- adjust Turn priority documentation
- Implement a transation based approach to modifying battle state

closes #51
indirectly fixes #49
closes #68 

This is one of 2 possible implementations that I created for logging changes in the battle. (See #68)

This takes a transaction based approach to implementing a battle log. This adds a new `Transaction`
interface that works somewhat similarly to how the `Turn` interface works. A `Transaction` represents
a change to the battle state, which can occur as a result of an agent's turn, or automatically (like
the burn status effect wearing off).

### Benefits

- It's more clear how to implement end-of-turn status effects.
- It's seperates state changes into single operations, which is probably easier for a user to process.
- There's no possibility that an Agent could submit a Turn that included a Result that it is not supposed to have.
- It's more clear how to indicate to the user that the battle has ended.

### Problems

- It makes the implementation of somewhat basic things a bit more complex and verbose.

### DIFF
--- a/battle_test.go
+++ b/battle_test.go
@@ -200,7 +200,7 @@ func TestPokemonSpeed(t *testing.T) {
 	pkmn1.Stats = [6]uint{30, 10, 10, 10, 10, 10}
 	party1.AddPokemon(&pkmn1)
 	party2 := NewParty(&a2, 1)
-	pkmn2 := GetPokemon(4)
+	pkmn2 := GetPokemon(7)
 	pkmn2.Moves[0] = &pound
 	pkmn2.Stats = [6]uint{30, 10, 10, 10, 10, 12}
 	party2.AddPokemon(&pkmn2)
--- a/battle_test.go
+++ b/battle_test.go
@@ -61,20 +61,20 @@ func TestBattleOneRound(t *testing.T) {
 			t.Fatalf("Must send out first pokemon in each at the beginning of the battle. Party %d gave: %v", p, got)
 		}
 	}
-	transations := b.SimulateRound()
-	if len(transations) != 2 {
-		t.Fatal("Expected only 2 transations to occur in a round")
+	transactions := b.SimulateRound()
+	if len(transactions) != 2 {
+		t.Fatal("Expected only 2 transactions to occur in a round")
 	}
 	logtest := []struct {
 		turn Transaction
 		want string
 	}{
 		{
-			turn: transations[0],
+			turn: transactions[0],
 			want: "Charmander used Pound on Squirtle for 3 damage.",
 		},
 		{
-			turn: transations[1],
+			turn: transactions[1],
 			want: "Squirtle used Pound on Charmander for 3 damage.",
 		},
 	}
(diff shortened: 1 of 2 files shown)


adapap (changes requested):
I think transaction should be changed to `BattleEvent`/`StateChange` as some interactions are more like events than exchanges. Referring to Transactions as BattleEvents below.

I like this implementation a lot more for the following reasons:
- Turn is separate from the BattleEvent as they serve two different purposes (one for a single Pokemon action, one for a change in the battle)
  - This is an important distinction for testing and implementation
- Easily connects/extends with existing turn interface (#68 also does this)

dyc3:
I picked the name Transaction because I felt like it more closely resembled an sql transaction:
> A transaction is a unit of work that is performed against a database.

For telling end users about weather effects and stuff, we can simply have noop informational transactions. When a pokemon loses all it's health, it should create another transaction that sets the pokemon as inactive, and sets the next pokemon as active, or creates another transaction to end the battle.

adapap:
A battle is not a database, but more like a network that clients connect to and interact with. I think a name like [Log](https://github.com/fonse/pokemon-battle/blob/master/src/battle.coffee), [Action](https://github.com/smogon/pokemon-showdown/blob/master/sim/battle-queue.ts) or Event is better. Additionally, the example in #12 shows that the string representation is dependent on which agent is reading through the events; therefore, the string function needs some reference to parties either through a function parameter or embedded in the action/log/event.

dyc3:
- `Log` doesn't seem appropriate because a `Transaction` can define how the state changes. 
- `Action` could be easily confused with `Turn`.
- `Event` is better, but I feel like the name implies the _cause_ of the state change, not the _effect_

> the example in #12 shows that the string representation is dependent on which agent is reading through the events

That's not really relevant because why would the agent care about the battle log in string form? It's not machine readable. The `BattleLog()` returns strings that are intended to be viewed by end users. Think of it like an announcer who is watching the battle. Making them agent-specific is a waste of time. There's no need to replicate every single tiny detail of the original game's UI, that's not what we are doing.

I think we have different ideas of what an Agent is.

battle_test.go
```diff
@@ -66,7 +66,30 @@ func TestBattleOneRound(t *testing.T) {
...
+		turn Transaction
+		want string
+	}{
+		{
+			turn: transations[0],
+			want: "Charmander used Pound on Squirtle for 3 damage.",
+		},
+		{
+			turn: transations[1],
+			want: "Squirtle used Pound on Charmander for 3 damage.",
```
carpenterd777, line 83: (nit)
```suggestion
	transactions := b.SimulateRound()
	if len(transactions) != 2 {
		t.Fatal("Expected only 2 transations to occur in a round")
	}
	logtest := []struct {
		turn Transaction
		want string
	}{
		{
			turn: transactions[0],
			want: "Charmander used Pound on Squirtle for 3 damage.",
		},
		{
			turn: transactions[1],
			want: "Squirtle used Pound on Charmander for 3 damage.",
```

carpenterd777 (approved):

PR #73, opened 2021-02-17
fix get opponents

### LINKED ISSUE #72: GetOpponents returns incorrect results
## Current Behavior
`GetOpponents` returns the same results as `GetAllies`

## Expected Behavior
`GetAllies` should return friendly targets.
`GetOpponents` should return enemy targets.

**Additional context**

dyc3:
- fix GetOpponents, fixes #72
- improve GetAllies and GetOpponents tests

### DIFF
--- a/battle.go
+++ b/battle.go
@@ -65,7 +65,7 @@ func (b *Battle) GetOpponents(p *party) []target {
 	opponents := make([]target, 0)
 	targets := b.GetTargets()
 	for _, target := range targets {
-		if target.Team == p.team {
+		if target.Team != p.team {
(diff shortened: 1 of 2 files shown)


battle.go
```diff
@@ -65,7 +65,7 @@ func (b *Battle) GetOpponents(p *party) []target {
 	opponents := make([]target, 0)
 	targets := b.GetTargets()
 	for _, target := range targets {
-		if target.Team == p.team {
+		if target.Team != p.team {
```
adapap, line 68: Woops :L

adapap (approved):

PR #79, opened 2021-02-17
bring back pokedex.go with codegen comment, fixes #77

### LINKED ISSUE #77: go generate does not regenerate pokedex_GEN.go
## Current Behavior
go generate does not regenerate pokedex_GEN.go

**To Reproduce**
1. delete `pokedex_GEN.go`
2. run `go generate`
3. See that `pokedex_GEN.go` is not created.

## Expected Behavior
`go generate` should regenerate `pokedex_GEN.go`, even if `pokedex_GEN.go` is not already present.

**Additional context**
Seems like `pokedex.go` was removed. `pokedex.go` should be recreated with 
```
//go:generate go run data/gen.go
```
_Note that there is no spaces in `//go:generate`_

dyc3:


carpenterd777 (approved):
does this allow the code to be generated without having the package be called main?

PR #81, opened 2021-02-18
Pokemon stat computation and quickly build parties

### LINKED ISSUE #61: Pokemon Party building tools
Users are going to want to build parties quickly, and we're going to want it to write tests.

Currently, the steps to building a party are:
1. Create the party
2. Create the pokemon
3. Set the pokemon moves
4. Set the pokemon stats
5. Set the pokemon's current hp to the max hp
6. Repeat 2-5 up to 5 more times
7. Add all the pokemon to the party

This kinda sucks, especially because of step 5.

There should be some way to generate pokemon with correct stats based on the national dex number and level

The process should look more like:
1. Create the party
2. Generate a pokemon
3. Set the pokemon moves
4. Repeat 2-3 up to 5 more times
5. Add all the pokemon to the party

However, it needs to still be possible to hand craft pokemon if needed. This would be just a convenience tool.

**Describe alternatives you've considered**

**Additional context**

carpenterd777:
- Start working on Pokemon constructor, add some failing tests
- Removed unimplemented fields from generation code, added nature table and methods, pokemon generation method, tests
- Move level to optional parameters
- Added functions to quickly make parties and swapped out functions in test with them

Closes #61


pokemon.go
```diff
@@ -53,6 +67,137 @@ func GetPokemon(natdex uint16) Pokemon {
...
+			return p
+		}
+	}
+
+	// Not exactly the best way to handle this
+	panic(fmt.Sprintf("unknown Pokedex number %v\n", natdex))
+}
+
+func computeLevelFromExp(exp uint, growth_rate int) uint8 {
+	experience_csv := getCsvReader("data/experience.csv")
```
dyc3, line 143: Reading from `data/` at runtime is not allowed. All info must be included in the compiled binary.

pokemon.go
```diff
@@ -53,6 +67,137 @@ func GetPokemon(natdex uint16) Pokemon {
...
+			}
+		}
+
+		lastExp = nextExp
+	}
+	panic("There was a problem computing the level.")
+}
+
+func computeExpFromLevel(level uint8, growth_rate int) uint {
+	experience_csv := getCsvReader("data/experience.csv")
```
dyc3, line 166: Reading from `data/` at runtime is not allowed. All info must be included in the compiled binary.

pokemon.go
```diff
@@ -53,6 +67,137 @@ func GetPokemon(natdex uint16) Pokemon {
...
+			if level == record_level {
+				return uint(parseInt(record[2]))
+			}
+		}
+	}
+	panic("There was a problem computing the experience.")
+}
+
+func getGrowthRateFromDexNumber(natdex uint16) int {
+	pokemon_species_csv := getCsvReader("data/pokemon_species.csv")
```
dyc3, line 186: Reading from `data/` at runtime is not allowed. All info must be included in the compiled binary.

dyc3 (changes requested):

pokemon.go
```diff
@@ -53,6 +67,137 @@ func GetPokemon(natdex uint16) Pokemon {
...
+			}
+			evs = param
+		case 4: // nature
+			param, ok := p.(*Nature)
+			if !ok {
+				panic(fmt.Sprintf("Parameter %d: expected type Nature, got %v", i, reflect.TypeOf(p)))
+			}
+			nature = param
+		}
+	}
```
adapap, line 124: Variadic interface for this case is really ambiguous for the user. Couple of solutions to make this more clear:
- `map[string]interface{}`: gives a name to each key rather than the index]
- custom struct type: good because has explicit field names that the user can see, but difficult to test for zero values (e.g. passing in 0 vs. not passing in any value)
- [further optional parameter techniques](https://stackoverflow.com/questions/2032149/optional-parameters-in-go): probably a bit too complex for our needs, but shows composability in Go

[This article](https://petomalina.medium.com/dealing-with-optional-parameters-in-go-9780f9bfbd1d) describes better approaches to optional parameters.

pokemon.go
```diff
@@ -53,6 +67,137 @@ func GetPokemon(natdex uint16) Pokemon {
...
+			p.computeStats()
+			return p
+		}
+	}
+
+	// Not exactly the best way to handle this
+	panic(fmt.Sprintf("unknown Pokedex number %v\n", natdex))
+}
+
+func computeLevelFromExp(exp uint, growth_rate int) uint8 {
```
adapap, line 142: Is this function necessary? Can we do `func (p *Pokemon) GainExperience(amount)` and `func (p *Pokemon) GainLevels(levels)` instead, with the growth tables being hard-coded in `pokedex_GEN.go`?

pokemon.go
```diff
@@ -78,6 +223,90 @@ func (p *Pokemon) HasValidEVs() bool {
...
+	for _, stat := range [5]int{STAT_ATK, STAT_DEF, STAT_SPATK, STAT_SPDEF, STAT_SPD} {
+		ev_term := math.Floor(float64(p.EVs[stat]) / 4)
+		base_iv_ev_level_term := (2*base_stats[stat] + uint(p.IVs[stat]) + uint(ev_term)) * uint(p.Level)
+		floor_term := math.Floor(float64(base_iv_ev_level_term) / 100)
+		nature_term := float64(floor_term+5) * natureModifiers[stat]
+		p.Stats[stat] = uint(math.Floor(nature_term))
+	}
+}
+
+func (p *Pokemon) Equals(other Pokemon) bool {
```
adapap, line 277: What's a scenario when we need to check if two Pokemon are equal, field-wise? You can have two Pokemon with the exact same stats, but we really should be checking pointers or index to see if they are the same.

adapap (commented):
Definitely see some ideas here. I think this will take some iteration before it's done.

pokemon.go
carpenterd777: I have some tests for `GeneratePokemon()` set up that use `Equals()` to ensure its working properly. I guess I could change the test for the constructor's accuracy to only compare the stats, but how should I verify that `GeneratePokemon()` is working as expected without a method like this?

adapap: Like you said, by the individual fields of the Pokemon. First write a test with no args and see if all the fields are their default value (level = 1, stats = 0 array, etc). Then, 1 parameter at a time you test that functionality: `GeneratePokemon(nature = adamant)` -> `if p.nature != adamant -> fail` and repeat.

dyc3: You should probably just be using `reflect.DeepEqual`.

pokemon_test.go
```diff
@@ -19,6 +19,120 @@ func TestPokemonName(t *testing.T) {
...
+				Stats:           [6]uint{20, 11, 10, 9, 10, 9},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(fmt.Sprintf("Pokemon constructor test"), func(t *testing.T) {
+			t.Parallel()
+			got := tt.pkmn
+			if !got.Equals(tt.want) {
```
dyc3, line 98:
```suggestion
			if !reflect.DeepEqual(got, tt.want) {
```

data/gen.go
```diff
@@ -93,6 +94,66 @@ func contains(s []int, e int) bool {
...
+	case 3:
+		return "STAT_SPATK"
+	case 4:
+		return "STAT_SPDEF"
+	case 5:
+		return "STAT_SPD"
+	default:
+		panic(fmt.Sprintf("Could not find stat %d", stat_id))
+	}
+}
```
dyc3, line 142: You could just make these maps

types.go
```diff
@@ -183,11 +183,77 @@ func (g Gender) String() string {
...
+		natureModifiers[n.StatUp] = 1
+	}
+
+	return natureModifiers
+}
+
+// implement Stringer
+
+func (n Nature) String() string {
+	return n.name
```
dyc3, line 258: Not covered in tests

dyc3 (changes requested):

pokemon.go
```diff
@@ -71,6 +130,115 @@ func (p *Pokemon) HasValidEVs() bool {
...
+		fmt.Sprintf("\tNature: %s\n", p.Nature) +
+		fmt.Sprintf("\tStats: %v\n", p.Stats) +
+		fmt.Sprintf("\tCurrent HP: %d\n", p.CurrentHP) +
+		fmt.Sprintf("\tHeld item: %v\n", p.HeldItem) +
+		fmt.Sprintf("\tMoves: %v\n", p.Moves) +
+		fmt.Sprintf("\tFriendship: %d\n", p.Friendship) +
+		fmt.Sprintf("\tOriginal trainer ID: %d\n", p.OriginalTrainerID) +
+		"}"
+	return outputString
+}
```
dyc3, line 240: What's this for? It's also not covered in tests

dyc3 (changes requested):

pokemon.go
carpenterd777: this mainly exists for use in tests

types.go
carpenterd777: whoops this shouldn't be here

dyc3 (approved):

PR #89, opened 2021-02-18
battle end

### LINKED ISSUE #74: A way to indicate that the battle has ended.
We need to be able to tell the user that the battle has ended. More specifically, we need a `Transaction` that will end the battle.

dyc3:
- add EndBattleTransaction, SimulateRound now also returns bool to indicate end of battle, update tests
- add TestBattleEndByKnockout

closes #74


battle.go
```diff
@@ -177,18 +177,24 @@ func (b *Battle) SimulateRound() []Transaction {
 					}
 				}
 				if !anyAlive {
```
carpenterd777, line 179: `anyAlive` is false when all of a *party*'s Pokemon have fainted right? shouldnt we instead check to ensure that all of one team's Pokemon have fainted? (ex. multi battle where Trainer 1 and Trainer 2 form one team, all of Trainer 1's pokemon are knocked out but Trainer 2 still has pokemon)

battle.go
```diff
@@ -85,7 +85,7 @@ func (b *Battle) Start() error {
 }
 
 // Simulates a single round of the battle.
```
carpenterd777, line 87:
```suggestion
// Simulates a single round of the battle. Returns processed transactions for this turn and indicates whether the battle has ended.
```

battle.go
dyc3: We can do that when we actually do double battles.

carpenterd777 (approved):

PR #107, opened 2021-02-19
chore: abstract transactions

### LINKED ISSUE #90: Abstract transactions for organization/testing
Transactions are becoming too complex to keep in the scope of battles. We can introduce a new file, `transactions.go` to manage the different types of transactions as they scale as well as test them separately without creating battle instances. Additionally, the logic performed by transactions should be separate so that we can unit test its functionality.

Quick example:
`ItemTurn` should add transactions based on the item function. Using a potion should produce an `ItemTransaction` (to consume the potion) and a `HealTransaction` (to restore HP). The handlers for both transactions should be separate functions that we can test.

adapap:
Transaction logic should be separate from its implementation (handling a transaction) and the battle (creating the transactions). This PR solves this by creating `transactions.go` to define the transactions, and creates helpers to test things like healing without setting up a new battle. In addition, some logic was changed to use fewer SLOC for readability.

Closes #90.


pokemon.go
```diff
@@ -101,10 +101,20 @@ func (p Pokemon) String() string {
...
+		p.CurrentHP -= amount
+	} else {
+		// prevent underflow
+		p.CurrentHP = 0
+	}
+}
+
 // Restore HP to a Pokemon. Can also be used to revive a fainted Pokemon.
-func (p *Pokemon) RestoreHP(amount uint) Transaction {
+func (p *Pokemon) Heal(amount uint) {
```
carpenterd777, line 115: oh its interesting that you can do both. with our current set up, would we be able to use a potion on a fainted Pokemon?

dyc3: probably

transactions_test.go
```diff
@@ -0,0 +1,3 @@
+package pokemonbattlelib
+
+// TODO: move separate tests for transactions here?
```
dyc3, line 3: delete this file, it's not necessary

battle.go
```diff
@@ -155,18 +151,12 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
...
 			case DamageTransaction:
-				if t.Target.CurrentHP >= t.Damage {
-					t.Target.CurrentHP -= t.Damage
-				} else {
-					// prevent underflow
-					t.Target.CurrentHP = 0
-				}
-				if t.Target.CurrentHP == 0 {
+				receiver := b.getPokemon(t.Target.party, t.Target.partySlot)
+				receiver.Damage(t.Damage)
```
dyc3, line 155: Not a huge fan of this, tbh. Particularly because the code it's replacing is extremely trivial, and because Damage is publicly facing.

dyc3 (commented):
I'm confused about what's happening in this PR, probably because I'm tired. This PR is making at least 3 distinct changes. Go smaller, it should be 1, maybe possibly 2.

Moving the transactions to another file is fine, the refactoring the transactions to use `target` instead is fine.

I'll review this again tomorrow, maybe it'll make more sense then

pokemon.go
carpenterd777: is that a fix that is relevant to this PR or should it go into a new issue?

battle.go
adapap: We can make things like heal, damage, cure, etc. be internal functions. The goal here is to make testing damage separate from testing damage transactions.

pokemon.go
carpenterd777: thought about it a bit, I think it can be resolved in its own issue

dcarpenter31 (approved):

battle.go
dyc3: Then we should separate the creation of transactions from the processing of transactions. All changes to the battle state should go through transactions anyway.

Here, we're just making the logic harder to read by requiring people to look in multiple places for it.

pokemon_test.go
```diff
@@ -19,6 +19,34 @@ func TestPokemonName(t *testing.T) {
...
+	}
+	p.heal(100)
+	if p.CurrentHP != 100 {
+		t.Errorf("expected Pokemon to have 100 HP, received %v", p.CurrentHP)
+	}
+}
+
+func TestDamagePokemon(t *testing.T) {
+	p := GetPokemon(1)
+	p.Stats[0] = 100
```
dyc3, line 38:
```suggestion
	p.Stats[STAT_HP] = 100
```

pokemon_test.go
```diff
@@ -19,6 +19,34 @@ func TestPokemonName(t *testing.T) {
 	}
 }
 
+func TestHealPokemon(t *testing.T) {
+	p := GetPokemon(1)
+	p.Stats[0] = 100
```
dyc3, line 24:
```suggestion
	p.Stats[STAT_HP] = 100
```

pokemon.go
```diff
@@ -79,10 +79,20 @@ func (p Pokemon) String() string {
...
+
 // Restore HP to a Pokemon. Can also be used to revive a fainted Pokemon.
-func (p *Pokemon) RestoreHP(amount uint) Transaction {
+func (p *Pokemon) heal(amount uint) {
 	if diff := p.Stats[STAT_HP] - p.CurrentHP; diff <= amount {
 		amount = diff
 	}
-	return HealTransaction{Target: p, Amount: amount}
+	p.CurrentHP += amount
 }
```
dyc3, line 98: Again, this logic is extremely trivial, and these functions are only used in one spot. All changes to battle state have to go through transactions anyway, so we should just test the whole transaction. I'd prefer if these functions didn't exist.

dyc3 (changes requested):

dyc3:
I'm going to try and cherry pick the first 2 commits and put them in a new PR.

dyc3 (changes requested):
**Stop adding new changes to this PR. It only makes it take longer to merge in.** Please, just fix the existing problems.

adapap:
Idk what's left here

pokemon.go
dyc3: Remove these, as indicated in previous review.

dyc3 (changes requested):

dyc3:
This is too hard to merge in now. I'll close this PR and open a new one.

PR #108, opened 2021-02-19
move helper types/constants to types.go

dyc3:


types.go
```diff
@@ -6,6 +6,49 @@ import (
...
+
+const (
+	InvalidGroup ExpGroup = iota
+	Erratic
+	Fast
+	MediumFast
+	MediumSlow
+	Slow
+	Fluctuating
+)
```
carpenterd777, line 49:
```suggestion
```
In #81 , I made this a set of `int` consts instead of a whole new type.

types.go
```diff
@@ -6,6 +6,49 @@ import (
...
+
+// Constants for IVs and EVs
+const (
+	MAX_FRIENDSHIP    = 255
+	MAX_EV            = 255
+	MAX_IV            = 31
+	MAX_STAT_MODIFIER = 6
+	MIN_STAT_MODIFIER = -6
+	TOTAL_EV          = 510
+)
```
carpenterd777, line 37: I feel like this should stay in `pokemon.go`, since these constants relate directly to the fields of that type. Maybe we can leave out `MAX_STAT_MODIFIER` and `MIN_STAT_MODIFIER`, but I think the rest should stay in `pokemon.go`

types.go
```diff
@@ -6,6 +6,49 @@ import (
 	"math/bits"
```
carpenterd777, line 6: if this file is meant to hold types relating to the fields of `Pokemon`, maybe this can be named `pokemon_field_types` or something more specific than just `types`

dyc3: Not necessarily. It's basically just types that don't have enough code to warrant having their own file.

carpenterd777 (approved):

PR #110, opened 2021-02-20
Refactor Transaction queue processing

dyc3:
This PR contains a small refactor to make it possible to add pre/post turn transactions. As a side effect, it probably makes testing transactions easier.﻿

### DIFF
--- a/battle.go
+++ b/battle.go
@@ -14,6 +14,9 @@ type Battle struct {
 	State    BattleState
 
 	parties []*party // All parties participating in the battle
+
+	tQueue     []Transaction
+	tProcessed []Transaction
 }
 
 type BattleState int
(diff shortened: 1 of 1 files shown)


battle.go
```diff
@@ -140,78 +141,93 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
...
+	if len(b.tQueue) > 0 {
+		log.Panic("FATAL: There are still unprocessed transactions at the end of the round.")
+	}
+	transactions := b.tProcessed
+	b.tProcessed = []Transaction{}
+	return transactions, b.State == BATTLE_END
+}
+
+// Add Transactions to the queue.
+func (b *Battle) QueueTransaction(t ...Transaction) {
```
carpenterd777, line 166: I really like this function, its basic but I def thinks it helps improve readability of the battle logic

carpenterd777 (approved):

PR #113, opened 2021-02-20
Implement EV Transactions

### LINKED ISSUE #92: EV Transactions
Add transactions and helper functions to manipulate EVs of a Pokemon.

adapap:
EV Transactions take a target, stat (e.g. STAT_DEF), and amount.

~Edit: Blocked by #150 for now.~
Closes #92

### DIFF
--- a/battle_test.go
+++ b/battle_test.go
@@ -689,7 +689,7 @@ var _ = Describe("Fainting", func() {
 		It("should switch to the next available Pokemon", func() {
 			Expect(b.Start()).To(Succeed())
 			t, _ := b.SimulateRound()
-			Expect(t).To(HaveLen(5))
+			Expect(t).To(HaveLen(6))
 			// Charmander smashed his nubby little fist into Squirtle as
 			// hard as he could. Spectators gasped and winced when the
 			// impact created a very audible crack. But it was not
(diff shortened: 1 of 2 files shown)


items.go
```diff
@@ -99,7 +101,21 @@ func (p *Pokemon) UseItem(i *Item) []Transaction {
...
+		t = append(t, HealTransaction{Target: p, Amount: 20})
 	}
 	return t
 }
+
+// Uses a vitamin which affects EVs and friendship
+func (p *Pokemon) UseVitamins(i *Item) (t []Transaction) {
+	// https://bulbapedia.bulbagarden.net/wiki/Friendship#In_Generation_IV
+	switch i.ID {
+	case ITEM_CALCIUM:
```
carpenterd777, line 113: do we want to implement vitamins in battle? in the original games, you cannot use vitamins in battle. (see https://bulbapedia.bulbagarden.net/wiki/Protein for an example) the only case in which a Pokemon can usually gain EVs in battle is by knocking out other Pokemon, though EV yield has not yet been generated from the data.

adapap: is there a list of battle-usable items? i thought I filtered the generated data to only allow those, but maybe it needs more filtering.

carpenterd777: yeah, there should be one in the miscellaneous section of the wiki

battle_test.go
```diff
@@ -22,11 +22,13 @@ func (a dumbAgent) Act(ctx *BattleContext) Turn {
...
+type itemAgent struct {
+	item int
+}
 
-// Always uses a potion on first Pokemon
-func (a healAgent) Act(ctx *BattleContext) Turn {
-	item := GetItem(ITEM_POTION)
+// Always uses the specified item on the first Pokemon
+func (a itemAgent) Act(ctx *BattleContext) Turn {
+	item := GetItem(a.item)
```
dyc3, line 31: Probably shouldn't change this, `healAgent` is used in other tests.

Use `RcAgent` instead

dyc3 (changes requested):

dyc3:
I think this PR is too out of date by now to merge in smoothly, and it's pretty small. Can we close it? @adapap

dyc3 (approved):

PR #117, opened 2021-02-20
Add Gomega and Ginkgo tests

### LINKED ISSUE #106: Refactor tests to use Gomega instead
See: https://onsi.github.io/gomega/

carpenterd777:
This also moves all the tests to its own package, which is apparently good practice for behavior driven design (BDD). Some unexported types needed to be exported as a result.

Closes #106


carpenterd777:
This refactor is kinda hefty, so if I could split the tests with someone that would be great

README.md
```diff
@@ -25,3 +25,10 @@ Using the github cli:
 ```bash
 gh pr create
 ```
+
+## Running Tests
+
+Ensure you have the ginkgo testing framework installed:
+```bash
+go install github.com/onsi/ginkgo/ginkgo
+```
```
dyc3, line 34: IIRC, go automatically grabs dependencies when you build anything

dyc3:
I would prefer if we could move stuff over in smaller chunks, one test file at a time. It'll be easier to manage the transition. This PR is fine to maintain what it has in it: the `pokemon_test` and `rng_test`, as it is fairly large already.

dyc3:
How do you run the tests? `go test` doesn't seem to work.

dyc3 (changes requested):
So far so good, but `go test` needs to work from the top level directory, and we need to be able to collect test coverage.

I also think that the test files should stay next to the things that they are testing, instead of going in it's own folder.

carpenterd777:
I don't think the files should stay in the top-level directory. Moving them into a subpackage not only increases organization of the project directory but ensures that we are testing only the behavior of the library instead of its implementation details.

Running the tests can be done atm by moving into the `pokemonbattlelib_test` directory and running `go test` or `ginkgo`. Is there a way to configure the tool we use for code coverage (I see it put comments on our prs)? It would be worth investigating if we can keep collecting test coverage but keep this directory structure

carpenterd777:
This might be useful:

https://onsi.github.io/ginkgo/#ginkgo-and-continuous-integration

dyc3:
Every other go project that I've seen puts tests next to the thing that they are testing. I see no reason to try and force something for organization if it's not easy, because it was clearly not intended to work that way. We'll probably be setting ourselves up for problems down the road.

carpenterd777:
Try running `ginkgo -r --randomizeAllSpecs --randomizeSuites --failOnPending --coverpkg=github.com/StevensSEC/pokemonbattlelib --trace --race --progress --coverprofile=coverage.out --outputdir=.` with this branch checked out

dyc3:
That works, but it's super verbose. It's going to be confusing/intimidating for beginners. I still think that we should still have the tests work with just `go test`

carpenterd777:
If its too verbose, just `ginkgo -r` will work, or `ginkgo -r --coverpkg=github.com/StevensSEC/pokemonbattlelib --coverprofile=coverage.out --outputdir=.`. Or we can use the vscode task again

pokemon.go
```diff
@@ -94,6 +96,19 @@ func WithNature(nature *Nature) GeneratePokemonOption {
...
+	}
+
+	var limited_moves [4]*Move
+	copy(limited_moves[:], moves)
+
+	return func(p *Pokemon) {
+		p.Moves = limited_moves
+	}
+}
+
```
dyc3, line 111: Should go in a different PR.

pokemon.go
```diff
@@ -173,9 +188,9 @@ func (p *Pokemon) GainExperience(exp int) {
 	remaining_exp := exp
 	toNextLevel := EXP_TABLE[p.GetGrowthRate()][int(p.Level+1)] - int(p.TotalExperience)
 
-	for remaining_exp > toNextLevel {
-		p.GainLevels(1)
+	for remaining_exp >= toNextLevel {
 		remaining_exp -= toNextLevel
+		p.GainLevels(1)
 		toNextLevel = EXP_TABLE[p.GetGrowthRate()][int(p.Level+1)] - int(p.TotalExperience)
 	}
```
dyc3, line 195: Bug fix, should go in a different PR.

dyc3 (changes requested):
Right now, basically all of our tests act as if they are a user of the package: by creating a battle that produces the conditions that it's testing for. The only exception I can think of off the top of my head is `apply`/`clear` for `StatusCondition`, and it's just there for the sanity check. 

If we _really, really_ want to write tests as if we are importing the package, we can do that, but the folder should be called `tests`, and we should not directly move our existing tests there.

If we decide to move different parts of the library to subpackages, it will be more cumbersome to dig through the tests folder to find the test file that we are looking for. It will be much easier if the test files are right next to the thing.

Additionally, as it stands with this implementation, this method of testing requires us to export functions that are not supposed to be exported. `getPokemon` on Battle and the `party` type were not originally intended to be exported.

I think that moving the tests to another folder is only going to cause us more pain down the line. Please put them back where they were.

Also, some tests have been commented out, which automatically makes this PR unmergable for me.

battle.go
```diff
@@ -35,12 +35,12 @@ func NewBattle() *Battle {
...
 
 // Adds one or more parties to a team in the battle
-func (b *Battle) AddParty(p ...*party) {
+func (b *Battle) AddParty(p ...*Party) {
 	b.parties = append(b.parties, p...)
 }
 
 // Gets a reference to a Pokemon using party ID and party slot
-func (b *Battle) getPokemon(party, slot int) *Pokemon {
+func (b *Battle) GetPokemon(party, slot int) *Pokemon {
```
dyc3, line 43: Revert this change
```suggestion
func (b *Battle) getPokemon(party, slot int) *Pokemon {
```

.vscode/tasks.json
```diff
@@ -6,13 +6,13 @@
 		{
 			"label": "Run Tests with Coverage",
 			"type": "shell",
-			"command": "go test -coverprofile=coverage.out -covermode=count && go tool cover -html=coverage.out",
+			"command": "ginkgo -r --coverpkg=github.com/StevensSEC/pokemonbattlelib --coverprofile=coverage.out --outputdir=. && go tool cover -html=coverage.out",
 			"windows": {
-				"command": "go test -coverprofile=\"coverage.out\" & go tool cover -html=\"coverage.out\""
+				"command": "ginkgo -r --coverpkg=github.com/StevensSEC/pokemonbattlelib --coverprofile=\"coverage.out\" --outputdir=\".\" & go tool cover -html=\"coverage.out\""
```
dyc3, line 11: Revert these changes

.github/workflows/test.yml
```diff
@@ -37,8 +37,14 @@ jobs:
...
 
+    - name: Get ginkgo
+      run: go get -v github.com/onsi/ginkgo/ginkgo
+
+    - name: Get gomega
+      run: go get -v github.com/onsi/gomega
+    
     - name: Test
-      run: go test -v -covermode=atomic -coverprofile=coverage.out ./...
+      run: ginkgo -r --randomizeAllSpecs --randomizeSuites --failOnPending --coverpkg=github.com/StevensSEC/pokemonbattlelib --trace --race --progress --coverprofile=coverage.out --outputdir=.
```
dyc3, line 47: Revert these changes

party.go
```diff
@@ -6,7 +6,7 @@ import (
 )
 
 // A Pokemon party. Can hold up to 6 Pokemon. Also manages how many pokemon are out on the battlefield.
-type party struct {
+type Party struct {
```
dyc3, line 9: Revert this change
```suggestion
type party struct {
```

dyc3 (changes requested):

dyc3 (approved):

PR #120, opened 2021-02-20
fix status condition check for non volatile statuses, adjust bits for status condition constants fixes #118

### LINKED ISSUE #118: Status Condition check returns false positives for non-volatile statuses
## Current Behavior
Status Condition check returns false positives for non-volatile statuses

**To Reproduce**
```go
StatusParalyze.check(StatusBurn) // returns true when it should be false.
```

## Expected Behavior
`check` should only return true for non-volatile statuses if all bits in the `NONVOLATILE_STATUS_MASK` are identical

**Additional Context**
Here's a function to get a string of all the status conditions and their bits
```go
func DebugStatusConditions() string {
	result := []string{}
	for c := StatusBurn; c <= StatusSleep; c++ {
		result = append(result, fmt.Sprintf("%-20s %032s", statusStrings[c], strconv.FormatUint(uint64(c), 2)))
	}
	for c := StatusBound; c <= StatusTorment; c <<= 1 {
		result = append(result, fmt.Sprintf("%-20s %032s", statusStrings[c], strconv.FormatUint(uint64(c), 2)))
	}
	return strings.Join(result, "\n")
}
```

dyc3:
This PR is blocking #119 until it is merged in.

### DIFF
--- a/types_test.go
+++ b/types_test.go
@@ -161,11 +161,11 @@ func TestStatusConditionCheck(t *testing.T) {
 	for ti, tt := range tests {
 		t.Run(fmt.Sprintf("Status Condition check: #%d", ti), func(t *testing.T) {
 			if tt.status.check(tt.check) != tt.want {
-				w := ""
+				messageMod := ""
 				if !tt.want {
-					w = " NOT"
+					messageMod = " NOT"
 				}
-				t.Errorf("Status Condition check failed %b, should%s match %b", tt.status, w, tt.check)
+				t.Errorf("Status Condition check failed %b, should%s match %b", tt.status, messageMod, tt.check)
 			}
 		})
 	}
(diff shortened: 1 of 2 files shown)


types_test.go
```diff
@@ -125,28 +125,47 @@ func TestStatusConditionCheck(t *testing.T) {
...
-			if !tt.status.check(tt.check) {
-				t.Errorf("Status Condition check failed %b, should have %b", tt.status, tt.check)
+	for ti, tt := range tests {
+		t.Run(fmt.Sprintf("Status Condition check: #%d", ti), func(t *testing.T) {
+			if tt.status.check(tt.check) != tt.want {
+				w := ""
+				if !tt.want {
+					w = " NOT"
+				}
+				t.Errorf("Status Condition check failed %b, should%s match %b", tt.status, w, tt.check)
```
carpenterd777, line 168:
```suggestion
				messageMod := ""
				if !tt.want {
					messageMod = " NOT"
				}
				t.Errorf("Status Condition check failed %b, should%s match %b", tt.status, messageMod, tt.check)
```

carpenterd777 (approved):

Issue #101, opened 2021-02-19
Non-volatile status conditions


dyc3:
Implement non-volatile status conditions and related transactions required for their effects.

- Burn
- Freeze
- Paralysis
- Poison
- Bad Poison
- Sleep

See: https://github.com/StevensSEC/pokemonbattlelib/wiki/Requirements#non-volatile-status-conditions

Partially dependent on #87

PR #122, opened 2021-02-20
add post-turn status effects: burn, poison

dyc3:
Part of #101


adapap (approved):
Outside the scope of this PR, would it not make more sense to have the methods for manipulating status effects on Pokemon rather than some random number type? `pokemon.CheckStatus(effect)` makes more sense to me.

dyc3:
That's really just syntactic sugar, IMO.

PR #124, opened 2021-02-21
add private rng field to battle

### LINKED ISSUE #87: Add rng provider to Battle.
We need some way for battles to generate random numbers predictably even if the user is running multiple battles at the same time.

To do this, we need to add a field to the battle struct with the type `RNG` (which is already defined in `rng.go`). 

https://github.com/StevensSEC/pokemonbattlelib/blob/8969e52b3cd7c9aaed7ebf3476f142ab454e1022/battle.go#L11-L17

`NewBattle` should initialize this field.

https://github.com/StevensSEC/pokemonbattlelib/blob/8969e52b3cd7c9aaed7ebf3476f142ab454e1022/battle.go#L28-L33

dyc3:
closes #87﻿

### DIFF
--- a/battle.go
+++ b/battle.go
@@ -2,6 +2,7 @@ package pokemonbattlelib
 
 import (
 	"log"
+	"math/rand"
 	"reflect"
 	"sort"
 )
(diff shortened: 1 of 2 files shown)


adapap (approved):
OK

A-very-Cunning-ham (approved):
Looks good to me

PR #125, opened 2021-02-21
add rcAgent, closes #75

### LINKED ISSUE #75: "Remote controlled" Agent for use in testing
We need some way to supply the battle with turns from an Agent in a specific order in tests.

An `Agent` that has a channel that it takes a `Turn` from and returns it when `Act` is called. This will allow us to write tests that check specific sequences of `Turns` easier.

**Describe alternatives you've considered**

**Additional context**

dyc3:
I didn't bother changing `TestRcAgent` to use the new testing framework because I'm tired and it's late.

### DIFF
--- a/battle_test.go
+++ b/battle_test.go
@@ -1,6 +1,9 @@
 package pokemonbattlelib
 
 import (
+	"fmt"
+	"testing"
+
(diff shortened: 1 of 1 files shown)


adapap (approved):
Weird but interesting.

PR #126, opened 2021-02-21
add ItemFlags, and update codegen

dyc3:


adapap (commented):
I think "usable in battle" might be enough to filter out the list of items that we can actually deal with, the other flags are definitely not necessary. Consumable can be a `bool` on the item struct. I'm going to assume the only time this distinction matters is when a held item is consumed, as things like bag/inventory are not part of this library.

dyc3:
> Consumable can be a bool on the item struct.

Yeah, but the whole point of doing it this way is so that you don't _have to_ have a bunch of bools on the struct. It doesn't make it any harder to check whether or not the condition is true, and it allows you to check more than one condition at a time.

> as things like bag/inventory are not part of this library

Yet ;)

adapap:
Okay, then the only change I would request is to not use string lookup, but create new constants for item flags. String keys are really hard to refactor.

adapap (approved):

Issue #127, opened 2021-02-21
FaintTransaction creates SendOutTransaction with incorrect partySlot


dyc3:
## Current Behavior
FaintTransaction creates SendOutTransaction with incorrect partySlot

## Expected Behavior
```go
partySlot: t.Target.partySlot,
```
should be
```go
partySlot: i,
```

Issue #84, opened 2021-02-18
Dead pokemon can still take turns.


dyc3:
## Current Behavior
If a pokemon faints before taking it's turn, it will ignore the fact that it is dead, and take it's turn anyway.

**To Reproduce**
Code snippet to reproduce the problem (if applicable)

<details>
<summary>Failing test case 1</summary>

```go
func TestBattleDeadPokemonForfeitTurn(t *testing.T) {
	a1 := Agent(dumbAgent{})
	a2 := Agent(dumbAgent{})
	party1 := NewParty(&a1, 0)
	pkmn1 := GetPokemon(4)
	pkmn1.Stats = [6]uint{30, 10, 10, 10, 10, 10}
	pkmn1.CurrentHP = 1
	pound := GetMove(1)
	pkmn1.Moves[0] = &pound
	party1.AddPokemon(&pkmn1)
	party2 := NewParty(&a2, 1)
	pkmn2 := GetPokemon(7)
	pkmn2.Stats = [6]uint{30, 10, 10, 10, 10, 100}
	pkmn2.CurrentHP = 30
	pkmn2.Moves[0] = &pound
	party2.AddPokemon(&pkmn2)
	b := NewBattle()
	b.AddParty(party1, party2)
	err := b.Start()
	if err != nil {
		t.Fatal("failed to start battle")
	}

	transactions, ended := b.SimulateRound()
	if !ended {
		t.Error("Expected SimulateRound to indicate that the battle has ended, but it did not.")
	}
	if n := len(transactions); n != 3 {
		t.Errorf("Expected 3 transactions to occur, got %d", n)
	}
	logtest := []struct {
		turn Transaction
		want string
	}{
		{
			turn: transactions[0],
			want: "Squirtle used Pound on Charmander for 3 damage.",
		},
		{
			turn: transactions[1],
			want: "Charmander fainted.",
		},
		{
			turn: transactions[2],
			want: "The battle has ended.",
		},
	}
	for _, tt := range logtest {
		got := tt.turn.BattleLog()
		if got != tt.want {
			t.Errorf("Expected battle log to be %s, got %s", tt.want, got)
		}
	}
}
```
</details>

<details>
<summary>Failing test case 2</summary>

```go
func TestBattleDeadPokemonForfeitTurn2(t *testing.T) {
	a1 := Agent(dumbAgent{})
	a2 := Agent(dumbAgent{})
	party1 := NewParty(&a1, 0)
[...]

PR #128, opened 2021-02-21
add a benchmark

dyc3:
The benchmark can be run with
```
go test -bench=.
```

The benchmark works, but some of the battles it creates are bugged because of #84., and #127

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -19,6 +19,14 @@
 			"program": "${workspaceFolder}"
 		},
 		{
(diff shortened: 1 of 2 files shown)


adapap (approved):
What does this do exactly? Also, `NewRandomParty` and `GetRandomMove` should probably be part of the party building suite.

dyc3:
@adapap  It should let us measure CPU and memory performance to some degree. I have no idea how useful it's going to be, but its something I'd like to try and keep track of.

PR #138, opened 2021-02-23
add move priority to sorting

### LINKED ISSUE #91: Add move priority to turn sorting
Sort turns based on move priority and add tests that specifically check for priority between turns and properly sorts them.

dyc3:
closes #91﻿


dyc3:
There seems to be a coverage problem. don't merge in just yet.

adapap (approved):

PR #144, opened 2021-02-24
Friendship Transactions

adapap:
Add Friendship Transactions and handle them in battle.

To Do:
- [x] Add tests for FriendshipTransaction
- [x] Add example code modifying friendship

Factors that affect friendship: https://bulbapedia.bulbagarden.net/wiki/Friendship#In_Generation_IV

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -7,9 +7,6 @@
 			"label": "Run Tests with Coverage",
 			"type": "shell",
 			"command": "go test -coverprofile=\"coverage.out\" && go tool cover -html=\"coverage.out\"",
-			"windows": {
-				"command": "go test -coverprofile=\"coverage.out\" & go tool cover -html=\"coverage.out\""
-			},
 			"problemMatcher": [
 				"$go"
 			]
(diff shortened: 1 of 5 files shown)


.vscode/tasks.json
```diff
@@ -7,12 +7,9 @@
 			"label": "Run Tests with Coverage",
 			"type": "shell",
 			"command": "go test -coverprofile=\"coverage.out\" && go tool cover -html=\"coverage.out\"",
-			"windows": {
-				"command": "go test -coverprofile=\"coverage.out\" & go tool cover -html=\"coverage.out\""
-			},
```
dyc3, line 12: any reason for removing this?

dyc3 (approved):

.vscode/tasks.json
adapap: Every time I ran that task it would not recreate the coverage output correctly, and it fails if there is no coverage file at all. So I end up using the linux version anyways - most people should install support for it if it's not already compatible. I also don't entirely know if it works sometimes, but that might just be me.

PR #151, opened 2021-02-24
add Paralysis, Freeze, Sleep status checks

### LINKED ISSUE #101: Non-volatile status conditions

dyc3:
closes #101﻿


battle.go
```diff
@@ -149,6 +149,42 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
...
+			// pre-move checks
+			forfeitTurn := false
+			if user.StatusEffects.check(StatusFreeze) || user.StatusEffects.check(StatusParalyze) {
+				var success bool
+				if user.StatusEffects.check(StatusFreeze) {
+					// TODO: use battle rng
+					success = b.rng.Get(1, 5) == 1
+				} else if user.StatusEffects.check(StatusParalyze) {
+					success = b.rng.Get(1, 4) != 1
+				}
```
adapap, line 162: Feels like this should be done in handling the transaction? Also, if we are not trying to mimic the original game, it seems pointless to have a transaction simply state that a Pokemon is frozen unless something else changes (such as forfeiting a turn or status wearing off).

transactions.go
```diff
@@ -96,3 +96,18 @@ func (t EndBattleTransaction) BattleLog() string {
...
+	Success      bool
+}
+
+func (t StatusEffectCheckTransaction) BattleLog() string {
+	failure := ""
+	if !t.Success {
+		failure = " and is unable to move"
+	}
+	return fmt.Sprintf("%s is %s%s.", t.Target.Pokemon.GetName(), t.Target.Pokemon.StatusEffects&NONVOLATILE_STATUS_MASK, failure)
+}
```
adapap, line 113: Think this would be better wrapped up as something like `ImmobilizeTransaction` where the cause is passed in as a `StatusEffect`. That way the same transaction can be used for moves/abilities that cause immobilization.

adapap (changes requested):
Don't think it's necessary to include `Success` in the transaction as that makes it try to mimic game logs too much, which we agreed is not the focus (e.g. heal transactions might not actually heal for their amount, hence a technically incorrect battle log).

transactions.go
dyc3: Yeah, I like that better.

battle.go
dyc3: Yeah, I was having a tough time figuring out how to do this, specifically because the turn can only be forfeited if it is a FightTurn. I'll rework it.

adapap (approved):

PR #153, opened 2021-02-25
Linting/CSV reduction

adapap:
Fix some linting errors with error checking, remove unused/redundant CSV. We can always re-add them if they become needed.


dyc3 (changes requested):
`version_groups.csv` is required

dyc3 (approved):

PR #154, opened 2021-02-26
Type Chart

adapap:
Changes the type effectiveness to use a simpler chart. Allows lookup in 1 table instead of three, and simplifies the computation.

### DIFF
--- a/types.go
+++ b/types.go
@@ -3,6 +3,7 @@ package pokemonbattlelib
 import (
 	"fmt"
 	"math"
+	"math/bits"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


dyc3 (changes requested):
I like the change to the type chart table thing, but not the change to GetElementalEffect because it requires adding in a loop. The current implementation is O(1), and requires significantly less map lookups on average.

dyc3 (changes requested):
Honestly, I'm not really feeling it. I'm probably biased, but I think that `GetElementalEffect` is easier to read as it is now.

Lets put this on the back burner for now.

dyc3:
I don't think we are going to merge this in. Are we good to close it @adapap ?

PR #164, opened 2021-02-27
add custom gomega matchers for transactions

### LINKED ISSUE #146: Chore(tests): Change transaction log checking
Every time a new transaction is added or modified, many tests need to be updated with the new information. This usually is derived from the order or number of transactions occurring in a given battle. Since it's not reasonable to limit the number of transactions that occur while expanding the library, I suggest that we take the approach of writing tests that assert that certain transactions _do_ or _do not_ occur.

For example, in testing fainting logic, we are not concerned with the specific damage that occurred, friendship lost, swapping Pokemon, exp gained, etc. We only want to test: does some transaction contain "X fainted."

This will simplify some tests as well and avoid redundancy.

dyc3:
This PR adds a couple of custom matchers to make testing transactions easier.

closes #146

### DIFF
--- a/util_test.go
+++ b/util_test.go
@@ -1,0 +1,158 @@
+package pokemonbattlelib
+
+import (
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/onsi/gomega/types"
+)
+
+// Gomega custom matchers, see: https://onsi.github.io/gomega/#adding-your-own-matchers
+
+// Used for custom gomega matchers. Checks to see if a is probably the same pokemon as b based on values that are unlikely to change as a result of a transaction.
+func comparePokemon(a, b *Pokemon) bool {
+	return a.NatDex == b.NatDex &&
+		a.Nature.name == b.Nature.name &&
+		a.Gender == b.Gender &&
+		a.Elemental == b.Elemental
+}
+
+// Used for custom gomega matchers. For simplicity, the fields of the struct are hardcoded. If we need to add more fields to `target` something is probably wrong.
+func compareTargets(a, b target) bool {
+	return comparePokemon(&a.Pokemon, &b.Pokemon) &&
+		a.party == b.party &&
+		a.partySlot == b.partySlot &&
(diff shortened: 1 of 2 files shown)


util_test.go
```diff
@@ -0,0 +1,158 @@
...
+
+// Gomega custom matchers, see: https://onsi.github.io/gomega/#adding-your-own-matchers
+
+// Used for custom gomega matchers. Checks to see if a is probably the same pokemon as b based on values that are unlikely to change as a result of a transaction.
+func comparePokemon(a, b *Pokemon) bool {
+	return a.NatDex == b.NatDex &&
+		a.Nature.name == b.Nature.name &&
+		a.Gender == b.Gender &&
+		a.Elemental == b.Elemental
+}
```
adapap, line 20: I guess we can't just compare a pointer here because copies are made? Is it worth having a global Pokemon ID/hash attached to the Pokemon to make this comparison safer?

dyc3: It would probably be easier (and faster) to just always refer to a pokemon by party/slot. Or we could change the `target` type so that we can have a pointer to the original pokemon, changing it after getting the `Turn` from the agent.

adapap: I think the first point makes sense, but I think the user should still be able to see the Pokemon without having to get it separately when making a turn. I don't think we should change pointers because that could lead to unexpected side effects, especially if something related to turns changes. Can't we just assume that if the party/slot match, the target is correct in the tests?

dyc3: > unexpected side effects, especially if something related to turns changes

Like what? If we just change the pointer back to the original copy, how would that affect other behavior?

> Can't we just assume that if the party/slot match, the target is correct in the tests?

Maybe? That feels kinda brittle to me, especially considering that we had #127.

adapap: I think I just need to see an implementation before judging this. Go with whatever your intuition is and it will probably be fine.

dyc3: Can we go ahead and merge this then?

adapap (approved):

PR #171, opened 2021-02-28
fix: change rng.Get usage to rng.Roll

adapap:


dyc3 (approved):

dyc3:
@adapap tests are failing with this one

battle.go
```diff
@@ -161,9 +161,9 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
 			if user.StatusEffects.check(StatusFreeze) || user.StatusEffects.check(StatusParalyze) {
 				var success bool
 				if user.StatusEffects.check(StatusFreeze) {
-					success = b.rng.Get(1, 5) == 1
+					success = b.rng.Roll(1, 5)
 				} else if user.StatusEffects.check(StatusParalyze) {
-					success = b.rng.Get(1, 4) != 1
+					success = b.rng.Roll(1, 4)
```
dyc3, line 166: Paralysis gives a 25% chance to fail taking the action, so chance of success should be 3/4.
```suggestion
					success = b.rng.Roll(3, 4)
```

dyc3 (changes requested):

battle.go
dyc3: I did a little testing with the RNG:
```go
	rng := LCRNG(982348932)
	results := map[int]int{}
	for i := uint(0); i <= 10000; i++ {
		n := rng.Get(1, 4)
		results[n]++
	}
	fmt.Printf("results: %#v", results)
```
Output:
```
results: map[int]int{1:3426, 2:3301, 3:3274}
```
That would probably be why it wasn't working before.

dyc3 (approved):
Re-approving because I don't feel like fixing RNG bugs rn

Issue #148, opened 2021-02-24
Battle/party setup utilities


adapap:
Quickly start battles in fewer lines of code:
- `NewRandomParty() party`
- `NewRandomBattle()` maybe...
- `NewBattleWithParties(parties ...*party)`
- `battle.SimulateRounds(rounds int)`
- `battle.SimulateUntilComplete()`

probably difficult and will require other utils
- `party.Reset()`
- `battle.Reset()`

PR #178, opened 2021-03-01
Add an optional HTTP API

dyc3:
- add http api framework, list moves endpoint
- add gitignore
- add generate endpoint

This PR adds an executable that serves HTTP requests. The goal is to create an interface that is easy to use in other programming languages, without requiring bindings to be created for each language.

### Justification
I need it for my SSW-215 project, and it would be useful to have in here anyway.

#### How to Run
```bash
go run ./http
```

### Other Details

In this scenario, Agents can be implemented with callback URLs (This has not been done yet).

I explored creating C bindings for other languages to consume, and it was very hard, and very buggy.

### TODO

- [x] Battle endpoints
- [x] Agent callbacks


http/agent.go
```diff
@@ -0,0 +1,64 @@
...
+		err = mapstructure.Decode(hT.Args, &t)
+		turn = Turn(t)
+	default:
+		panic("unknown turn number")
+	}
+	if err != nil {
+		panic(err)
+	}
+	return turn
+}
```
adapap, line 64: (nit) I feel like there's a better way to represent the type of a turn over an `int`.

http/api.go
```diff
@@ -0,0 +1,175 @@
...
+
+func buildRouter() *http.ServeMux {
+	mux := http.NewServeMux()
+	mux.HandleFunc("/pokedex/list/moves", HandleListMoves)
+	mux.HandleFunc("/pokedex/generate", HandleGeneratePokemon)
+	mux.HandleFunc("/battle/new", HandleCreateBattle)
+	mux.HandleFunc("/battle/simulate", HandleBattleSimulate)
+	mux.HandleFunc("/agent/dumb", HandleDumbAgent)
+	return mux
+}
```
adapap, line 21: I've pretty much exclusively used Gin for web serving in Go. Just handles a lot more out of the box if you want something more sophisticated.

http/api.go
```diff
@@ -0,0 +1,175 @@
...
+	Name string
+}
+
+// Convert a Pokemon into an httpPokemon, suitable for JSON marshalling for HTTP responses that contain pokemon.
+func convertPokemon(p *Pokemon) *httpPokemon {
+	return &httpPokemon{
+		Pokemon: *p,
+		Name:    p.GetName(),
+	}
+}
```
adapap, line 44: This kind of pattern should be avoided IMO. Add JSON tags to relevant structs [shown here](https://gobyexample.com/json).

http/api.go
```diff
@@ -0,0 +1,175 @@
...
+	w.Write(bytes)
+}
+
+var nextBattleId int
+var battles = map[int]*Battle{}
+
+type newBattleArgs struct {
+	Parties      [][]*Pokemon
+	CallbackUrls []string
+}
```
adapap, line 115: [jsoniter](https://github.com/json-iterator/go) has been fantastic to avoid having to unmarshal responses into a specific struct. It allows you to extract fields dynamically and check for them by key, with type checking as well.

http/util.go
```diff
@@ -0,0 +1,42 @@
...
+func randValues(max int) [6]uint8 {
+	return [6]uint8{
+		uint8(rand.Intn(max)),
+		uint8(rand.Intn(max)),
+		uint8(rand.Intn(max)),
+		uint8(rand.Intn(max)),
+		uint8(rand.Intn(max)),
+		uint8(rand.Intn(max)),
+	}
+}
```
adapap, line 20: This might as well be `NewRandomParty` (#148)

adapap (commented):
Probably a good start, but doesn't feel like Go code in some ways

http/agent.go
dyc3: I tried, but the type does not get preserved when marshaling to JSON and back, so we have to to it manually. This was definitely the easiest way.

http/api.go
dyc3: But we need the name, which is not present on `Pokemon`.

http/api.go
dyc3: I'm trying to avoid adding dependencies as much as possible.

http/util.go
dyc3: Yeah, it's a placeholder that needs to be replaced later.

http/api.go
dyc3: Again, I'm trying to avoid adding dependencies as much as possible.

http/api.go
adapap: You can write a custom `MarshalJSON` and `UnmarshalJSON` method on the original struct which can have custom effects. See [this article](http://choly.ca/post/go-json-marshalling/) for a simple example.

dyc3: Ah, I didn't think to introduce another struct to marshal the data

dyc3: Ok, so it turns out that you can't implement methods on types that you don't define, and since `Pokemon` is not defined in this package, I can't implement `MarshalJSON`/`UnmarshalJSON`

adapap: I don't see any problem with defining those methods in the original library, even if it's in a separate file/in `util.go`. Can even make it take a function so that it's customizable such as `p.MarshalJSON(customHandlerFunc)`. Also, have you looked at [WASM](https://www.sitepen.com/blog/compiling-go-to-webassembly)?

dyc3: I don't see the relavance of web assembly, just seems like extra steps.

dyc3: Also, if we change the function signature of `MarshalJSON` then we don't implement the `Marshaler` interface.

adapap (approved):
LGTM - I can't imagine making the frontend for this D:

PR #181, opened 2021-03-02
fix: status clear

### LINKED ISSUE #174: StatusCondition.clear() has weird behavior.
## Current Behavior
Using `clear()` on a `StatusCondition` can be used to apply a status condition if it is not present. It can also be used to give it a different non-volatile status condition entirely, in a way that doesn't make any sense.

**To Reproduce**
```go
s := StatusNone
s.clear(StatusBurn)
if s == StatusBurn {
    panic("should not happen")
}
```

```go
s := StatusNone
s.clear(StatusConfusion)
if s == StatusConfusion {
    panic("should not happen")
}
```

```go
s := StatusParalyze
s.clear(StatusFreeze)
if s == StatusBurn {
    panic("should not happen")
}
```

## Expected Behavior
`clear` should only xor status conditions that are already present.

**Additional context**

adapap:
Simple enough? Closes #174

### DIFF
--- a/types_test.go
+++ b/types_test.go
@@ -239,6 +239,30 @@ func TestStatusConditionClear(t *testing.T) {
 			clear: StatusBound,
 			want:  StatusSleep | StatusLeechSeed,
 		},
+		{
+			name:  "Should not clear status that does not exist",
+			from:  StatusNone,
+			clear: StatusBurn,
+			want:  StatusNone,
+		},
+		{
+			name:  "Should not clear freeze",
+			from:  StatusParalyze,
+			clear: StatusConfusion,
+			want:  StatusParalyze,
+		},
+		{
+			name:  "Should clear only existing effects",
+			from:  StatusParalyze | StatusBound | StatusCantEscape,
+			clear: StatusConfusion | StatusBound | StatusCantEscape,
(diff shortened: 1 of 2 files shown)


dyc3 (changes requested):
Doesn't have any tests. The solution is definitely not that simple.

types_test.go
```diff
@@ -239,6 +239,30 @@ func TestStatusConditionClear(t *testing.T) {
...
+		{
+			name:  "Should not clear status that does not exist",
+			from:  StatusNone,
+			clear: StatusBurn,
+			want:  StatusNone,
+		},
+		{
+			name:  "Should not clear freeze",
+			from:  StatusParalyze,
+			clear: StatusConfusion,
```
dyc3, line 251:
```suggestion
			clear: StatusFreeze,
```

dyc3 (changes requested):

dyc3 (changes requested):
I fixed the test case for you. The solution still does not fix the bug.

adapap:
I decided to simplify the logic instead of making the status condition check more complex. It's a lot more maintainable this way IMO.

dyc3 (approved):

PR #185, opened 2021-03-02
Pokemon Constants

adapap:
Since we have constants for items and moves, and they are rather useful, I don't see a compelling reason not to have some for Pokemon as well.

Waiting for #184


dyc3 (changes requested):
Constants should look like `PkmnBulbasaur` instead of `PkmnBULBASAUR`

dyc3 (approved):

PR #187, opened 2021-03-03
consume item only if it is consumable, move UseItem to ItemTransaction mutation

dyc3:


adapap (approved):
LGTM - Tests can go in another PR? #188 perhaps

PR #188, opened 2021-03-03
add in-a-pinch berries

### LINKED ISSUE #166: Add In-a-pinch Berries
Add functionality for each of these berries in battle.

### In-a-pinch
Num | Item | Effect
-- | -- | --
57 | Apicot Berry | Held: Consumed at 1/4 max HP to boost Special Defense.
54 | Ganlon Berry | Held: Consumed at 1/4 max HP to boost Defense.
58 | Lansat Berry | Held: Consumed at 1/4 max HP to boost critical hit ratio by two stages.
53 | Liechi Berry | Held: Consumed at 1/4 max HP to boost Attack.
56 | Petaya Berry | Held: Consumed at 1/4 max HP to boost Special Attack.
55 | Salac Berry | Held: Consumed at 1/4 max HP to boost Speed.

This issue was split off from #47.

dyc3:
Closes #166

### DIFF
--- a/battle_test.go
+++ b/battle_test.go
@@ -2,6 +2,7 @@ package pokemonbattlelib
 
 import (
 	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 )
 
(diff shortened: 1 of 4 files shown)


battle_test.go
```diff
@@ -1062,3 +1063,61 @@ var _ = Describe("Status Conditions", func() {
...
+				GetMove(MoveBite),
+			),
+			WithIVs([6]uint8{1, 1, 1, 20, 1, 1}),
+		)
+		i := GetItem(itemId)
+		holder.HeldItem = &i
+		holder.CurrentHP = holder.Stats[StatHP] / 2
+		p2 := NewOccupiedParty(&a2, 1, holder)
+		b := NewBattle()
+		b.SetSeed(1337)
```
adapap, line 1095: Please use `TestRNG` or one of the presets instead of a seed here.

items.go
```diff
@@ -127,14 +127,47 @@ func (p *Pokemon) UseMedicine(i *Item) (t []Transaction) {
...
+			Stages: 1,
+		})
 	case ItemSalacBerry:
+		t = append(t, ModifyStatTransaction{
+			Target: p,
+			Stat:   StatSpeed,
+			Stages: 1,
+		})
 	case ItemStarfBerry:
+		// TODO: boost random stat, requires battle RNG to be available.
```
adapap, line 170: Do you have ideas on whether we should pass in the battle object, BattleContext, or something else?

adapap (changes requested):

items.go
dyc3: Honestly, I'd prefer to move it all back into the transaction's `Mutate()`

battle_test.go
dyc3: This should be a quick fix, but changing the rng actually causing the test to be _more_ brittle.

items.go
adapap: I agree, I had implemented that system for items before the transaction system was well defined.

adapap (approved):

Issue #189, opened 2021-03-03
Getters for Pokemon stats


adapap:
While most items/moves affect stats by a number of stages from -6 to +6, certain items and abilities also multiply stats by percentages. We need a way to keep track of these multipliers, their targeted stats, and durations.

Something like this:
```
func (p *Pokemon) GetAttack(b *Battle) int {
  // base attack
  // modify from multipliers
  // modify from stages
  // special conditions?
}
```

Relevant:
#166 
#135 
#58 
#96

PR #158, opened 2021-02-26
Elemental type modifier

SWSoftware1044:
Adds battle modifier for Elemental Type matchups.

closes #85

PR #198, opened 2021-03-05
Miscellaneous Items

adapap:
Implementation of [misc items](https://veekun.com/dex/items/misc).

# Bad Held Items
- [x] Flame Orb
- [x] Full Incense
- [x] Iron Ball
- [x] Lagging Tail
- [x] Sticky Barb
- [x] Toxic Orb

# Choice Items
- [x] Choice Band
- [x] Choice Scarf
- [x] Choice Specs

## Blockers:
~#158 (iron ball changes immunities)~
~#189 (iron ball changes speed calculation, choice items affect stats)~

* Requires using items to have reference to battle
  * Perhaps a better way to get a Pokemon as a target? Like keeping track of the pointer and changing it (see [here](https://github.com/StevensSEC/pokemonbattlelib/pull/164#discussion_r584187419))


transactions.go
```diff
@@ -227,6 +240,16 @@ func (t PPTransaction) Mutate(b *Battle) {
 	}
 }
 
+// A transaction to change the held item of a Pokemon
+type SwapItemTransaction struct {
```
dyc3, line 244: (nit) Should probably be `GiveItemTransaction`

types.go
```diff
@@ -153,14 +153,17 @@ var doubleEffect = map[Type]Type{
...
 	reduce := bits.OnesCount32(uint32(halfEffect[move] & def))
 	increase := bits.OnesCount32(uint32(doubleEffect[move] & def))
 	effect := increase - reduce
+	if noEffect[move]&def > 0 {
+		if move&TypeGround > 0 && def&TypeFlying > 0 && item == ItemIronBall {
+			// Flying immunity negated
+		} else {
+			return NoEffect
+		}
+	}
```
dyc3, line 166: This should be handled in the `EffectiveType` getter method for the pokemon

pokemon.go
```diff
@@ -308,9 +313,16 @@ func (p *Pokemon) SpecialDefense() uint {
...
-	return uint(effective)
+	effective := p.Stats[StatSpeed]
+	if p.HeldItem != ItemNone {
+		switch p.HeldItem {
+		case ItemIronBall:
+			effective /= 2
+		case ItemChoiceScarf:
+			effective = (effective * 150) / 100
+		}
+	}
```
dyc3, line 324: No need for the extra `if`, just having the `switch` there would be equivalent.

battle.go
```diff
@@ -160,6 +164,10 @@ func (b *Battle) sortTurns(turns *[]TurnContext) {
 				if mvA.Priority() != mvB.Priority() {
 					return mvA.Priority() > mvB.Priority()
 				}
+				// Held item priority
+				if v, ok := pkmnA.metadata[MetaPriorityLast].(bool); ok && v {
+					return false
+				}
```
dyc3, line 170: Why can't you just check for the item being held?

dyc3 (changes requested):

types.go
adapap: I don't think this works because we can't change the type of the Pokemon in this case, we only want to remove it's immunity. If we change the Flying type to something like normal, then we introduce a super-effective Fighting matchup that is not expected.

battle.go
adapap: I suppose we can. I figured this would be a common enough case that other things can use the same meta tag to have it handled in the current/next round.

dyc3: I want to avoid using `metadata` as much as possible. It should only be used when no other option is possible.

types.go
dyc3: I see. I would still prefer this implementation not modify `GetElementalEffect`. Looking at bulbapedia, holding an iron ball applies [Grounded](https://bulbapedia.bulbagarden.net/wiki/Grounded) to the pokemon. We should have a helper function to determine if a pokemon is grounded. Then, we should have a special condition in `CalcMoveDamage` to handle that case, another helper function to handle all the special cases around type effectiveness.

dyc3 (approved):

PR #199, opened 2021-03-05
feat: add getters for all pokemon stats

### LINKED ISSUE #189: Getters for Pokemon stats

adapap:
Needed for a couple of PRs, closes #189

### DIFF
--- a/pokemon.go
+++ b/pokemon.go
@@ -230,32 +230,25 @@ func (p *Pokemon) computeStats() {
 	if !p.HasValidEVs() {
 		panic(fmt.Sprintf("Failed to compute stats for %s, evs %d invalid", p.GetName(), p.EVs))
 	}
-	// Set stats given base values (IV, EV, Level, Nature)
-	baseStats := p.GetBaseStats()
-	// Calculate base for HP
+
+	// get base stats
+	base_stats := p.GetBaseStats()
+
+	// compute HP
 	hp_ev_term := math.Floor(float64(p.EVs[StatHP]) / 4)
-	base_iv_ev_level_hp_term := (2*uint(baseStats[StatHP]) + uint(p.IVs[StatHP]) + uint(hp_ev_term)) * uint(p.Level)
+	base_iv_ev_level_hp_term := (2*uint(base_stats[StatHP]) + uint(p.IVs[StatHP]) + uint(hp_ev_term)) * uint(p.Level)
 	hp_floor_term := math.Floor(float64(base_iv_ev_level_hp_term) / 100)
-	effective := uint(hp_floor_term + float64(p.Level) + 10)
-	p.Stats[StatHP] = effective
-	p.CurrentHP = effective
-	// Calculate base for all other stats
-	for _, stat := range []int{StatAtk, StatDef, StatSpAtk, StatSpDef, StatSpeed} {
-		base := float64(baseStats[stat])
-		// Other stat formula
+	p.Stats[StatHP] = uint(hp_floor_term + float64(p.Level) + 10)
+	p.CurrentHP = p.Stats[StatHP]
+
+	// compute all other stats
+	natureModifiers := p.Nature.getNatureModifers()
+	for _, stat := range [5]int{StatAtk, StatDef, StatSpAtk, StatSpDef, StatSpeed} {
 		ev_term := math.Floor(float64(p.EVs[stat]) / 4)
-		base_iv_ev_level_term := (2*uint(base) + uint(p.IVs[stat]) + uint(ev_term)) * uint(p.Level)
-		base = math.Floor(float64(base_iv_ev_level_term)/100 + 5)
-		n := p.Nature
-		if n.StatUp != n.StatDown {
-			if n.StatUp == stat {
-				base *= 1.1
-			}
-			if n.StatDown == stat {
-				base *= 0.9
-			}
-		}
-		p.Stats[stat] = uint(base)
+		base_iv_ev_level_term := (2*uint(base_stats[stat]) + uint(p.IVs[stat]) + uint(ev_term)) * uint(p.Level)
+		floor_term := math.Floor(float64(base_iv_ev_level_term) / 100)
+		nature_term := float64(floor_term+5) * natureModifiers[stat]
+		p.Stats[stat] = uint(math.Floor(nature_term))
 	}
 }
 
--- a/pokemon.go
+++ b/pokemon.go
@@ -35,11 +35,6 @@ type PokemonMeta int
 
 const (
 	MetaLastMove PokemonMeta = iota
-	MetaUpdateAttack
-	MetaUpdateDefense
-	MetaUpdateSpecialAttack
-	MetaUpdateSpecialDefense
-	MetaUpdateSpeed
 )
 
 // Keeps track of base stats and EV yield for a Pokemon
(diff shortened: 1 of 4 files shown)


dyc3:
Why do we need these? We already denied a similar PR a couple of weeks ago.

adapap:
Otherwise our logic will blow up for calculating stats since base stats differ from the effective stat used in many calculations. See #189, but essentially we need to account for:
- Item modifiers
- Metadata that affects stats
- The multiplier from other stages
- Moves that affect stats

Every time we reference `p.Stats[Stat___]`, we are not accounting for any of those components that are being introduced to the library. This PR does not include setters for stats which was part of #58, but I think getters or some way to reuse this computation is necessary.

dyc3:
That's what `computeStats` is for. We just run that whenever the stats get invalidated.

adapap:
I thought compute stats would only be for determining stats when generating a new Pokemon. That would get super messy if we just kept all the logic in one function, and its unnecessary to constantly re-compute every stat which would be needed if we didn't have specific getters.

I feel like stats should be, essentially, base stats + IVs + nature (unchanging statistics). Effective stats can be calculated separately.

dyc3:
It's going to get messy anyway because there are so many edge cases, it's inevitable. `computeStats` doesn't have to be computed all the time, only when a state change occurs that requires `Stats` to be updated.

Now, it might be good to have getters/setters as a wrapper around this behavior, but it's probably worse for performance if we calculate stats every single time we get the stat.

adapap:
> Now, it might be good to have getters/setters as a wrapper around this behavior, but it's probably worse for performance if we calculate stats every single time we get the stat.

What do you mean by this? As in call something like `p.getHP()` inside `computeStats`?

dyc3:
No, setters around things that affect stats so you can mark `Stats` as dirty when necessary. When the stats are dirty, run `computeStats` and mark it as clean. This would definitely get complex pretty fast though.

dyc3:
Are there any scenarios where the value of a stat changes that's not accounted for by checking IVs/EVs, `StatModifiers`, the held item, and nature?

adapap:
[Abilities](https://bulbapedia.bulbagarden.net/wiki/Download_(Ability)) and a ton of [moves](https://bulbapedia.bulbagarden.net/wiki/Category:Moves_by_stat_modification), probably more

dyc3:
I don't see anything that we can't already handle with a `ModifyStatTransaction`.

adapap:
I guess not. Either way, there needs to be a place where base stats are mixed with multipliers from items and stages from stat modifiers, so there's float arithmetic happening. My goal is just to make it easier to get the effective stat value, and keeping it organized when changing it.

I just pushed a draft to try to organize the logic. It's hard to visualize without every other edge case, but now it's divided by stat so it might be cleaner.

pokemon.go
```diff
@@ -219,39 +219,108 @@ func (p *Pokemon) computeStats() {
...
 
-// implement Stringer
+func (p *Pokemon) setSpeed(base int) {
+	effective := p.getOtherStat(base, StatSpeed)
+	// Set effective stat
+	p.Stats[StatSpeed] = uint(effective)
+}
+
+// Stat getters (aliases)
+func (p *Pokemon) HP() uint {
```
dyc3, line 291: To not be confused with `CurrentHP`
```suggestion
func (p *Pokemon) MaxHP() uint {
```

pokemon.go
```diff
@@ -219,39 +219,108 @@ func (p *Pokemon) computeStats() {
...
+	// Set effective stat
+	p.Stats[StatSpDef] = uint(effective)
 }
 
-// implement Stringer
+func (p *Pokemon) setSpeed(base int) {
+	effective := p.getOtherStat(base, StatSpeed)
+	// Set effective stat
+	p.Stats[StatSpeed] = uint(effective)
+}
```
dyc3, line 288: I think `update` instead of `set` is more clear. Why can't we just leave `computeStats` the way it was?

Also, there is a bunch of repeated code.

dyc3 (approved):

dyc3 (changes requested):
misclick

pokemon.go
adapap: Agree with update, will change. As for changing `computeStats`, I just wanted to have effects on stats be separated more than just blocks in one function, as well as separating base stat calculation from things that are dynamic like items/abilities/modifiers. I chose to bring it out into different setters because it tightly couples implementation with that function, but it has its pros/cons. 

I also agree about repeating some of the operations, I don't really have a clear idea of what the best way of doing this is yet.

adapap:
I think this commit better demonstrates what I wanted to achieve. It separates base stat calculation from modifiers that come during battles.

pokemon.go
```diff
@@ -230,39 +230,115 @@ func (p *Pokemon) computeStats() {
...
-		ev_term := math.Floor(float64(p.EVs[stat]) / 4)
-		base_iv_ev_level_term := (2*uint(base_stats[stat]) + uint(p.IVs[stat]) + uint(ev_term)) * uint(p.Level)
-		floor_term := math.Floor(float64(base_iv_ev_level_term) / 100)
-		nature_term := float64(floor_term+5) * natureModifiers[stat]
-		p.Stats[stat] = uint(math.Floor(nature_term))
-	}
+func (p *Pokemon) updateAttack(base float64) {
+	effective := base
+	// TODO: attack modifiers
+	p.Stats[StatAtk] = uint(effective)
```
carpenterd777, line 281: my instinctual thought is to be wary of overwriting the stats field with effective stats. this is because I'm afraid that we'll encounter a move or ability whose effect scales with a Pokemon's base stats rather than their effective stats, though I cannot think of any off of the top of my head.

what if we stored the effective stats in another field?

carpenterd777: or we could use a method to compute the effective stats and return an effective stat array

dyc3: If we need the base stats, can't we just grab them again? Shouldn't `Stats` _be_ the effective stats though?

dyc3:
I've done a little bit of research, and it looks like specific stat modifications are calculated whereever the stats are used, so the damage calculation for example. So the only reason `Stats` should change is when the pokemon levels up.

adapap:
I don't think there's a point to not keeping the getters. We are always going to have to recompute stats, and there should be an easy way to do so for each stat. If we go back to the idea of marking stats that need to be recomputed, we can do something like this:

```golang
func (p *Pokemon) Attack() uint {
  if p.metadata[MetaUpdateAttack] == true {
    p.metadata[MetaUpdateAttack] = false
    // calculate effective attack
    p.Stats[StatAtk] = effective
  }
  return p.Stats[StatAtk]
}
```

This way, we can achieve both goals:
- Stats are computed "as needed" by setting the relevant meta tags
- We use getters so that we are sure that the stat we need is the latest version of the effective stat, and the logic is abstracted from battles

When a Pokemon levels up (or somehow otherwise changes their base stats), `p.computeStats()` is called which updates those base values, and all of the meta tags can be set.

pokemon.go
adapap: Agree here that stats is the effective stats. This PR is concerned with the way the effective stats are calculated/updated. Base stats are always available from `p.GetBaseStats()`

dyc3:
But aren't the stats used in only one location? Would it not make sense to put those calculations where they are being used? I guess it would be useful for Agents to not need to calculate all the edge cases themselves.

Looking at it again, I think we should leave `Stats` the way it was before, and just have the getters compute the effective values using `Stats`, and not updating it.

adapap:
> But aren't the stats used in only one location?

Stats will be needed at least 3 times with the pre-round, mid-round, and post-round checks. Not including the edge cases that merit other checks.

> Looking at it again, I think we should leave `Stats` the way it was before, and just have the getters compute the effective values using `Stats`, and not updating it.

I'll try to do this. Guess we don't need the recompute flags in this case.

dyc3 (approved):

PR #200, opened 2021-03-05
feat: add ev yield data (create PokemonBaseStats)

### LINKED ISSUE #150: EV Yield Data
Add EV Yield for defeating specific Pokemon to generated data.

Suggestion:
```golang
type EVYield struct {
    stat int
    amount int
}
// ...
{
  pokemon.Yield EVYield
}
```

adapap:
Created `PokemonBaseStats` as a wrapper for base stat values and EV yield. Closes #150


adapap:
Don't understand why linter fails, @dyc3?

dyc3:
> Make sure generated code is up to date

run `go generate` and commit

adapap:
Must be some tool version issue. I will try to fix but this works fine for me locally

data/gen.go
```diff
@@ -577,20 +589,18 @@ func main() {
...
-
-		base_stat_array[dex_num][stat_id] = stat_value
+		ev := parseInt(record[3])
+		baseStats[dex_num].stats[stat_id] = stat_value
+		baseStats[dex_num].ev[stat_id] = ev
 	}
 
-	for dex_num, stats := range base_stat_array {
-
+	for dex_num, v := range baseStats {
```
dyc3, line 597: It looks like its swapping around the order of the items in `pokemonBaseStats`. The iteration order of a map is not guaranteed by the go compiler.

Change this so that it iterates according to the length of the map, so something like
```go
for i := 1; i < len(baseStats); i++ {
```

dyc3 (changes requested):

dyc3 (approved):

PR #205, opened 2021-03-06
Refactor how Nature works

### LINKED ISSUE #99: Table of Natures
We need a constant table of `Nature`s for users to be able to assign to Pokemon. See https://github.com/StevensSEC/pokemonbattlelib/wiki/Requirements-(Miscellaneous)#all-natures-and-the-stats-they-affect

dyc3:
- add natures.csv back in
- add codegen for Nature, change Nature type to uint8

This PR makes 2 significant changes to how the type `Nature` works:
1. Change `Nature` to be a uint8, because that is all the information that is necessary to derive other values.
2. Use codegen to generate Nature constants, their stat modifiers, and their string representation.

This also has the side effect of implementing the rest of the natures, and `GetNature` is no longer required.

`getNatureModifers()` was left intact to make merging other currently open PRs less agonizing, but marked as deprecated, and should not be used in new PRs.

closes #99


adapap (commented):
I like this change but do we need to have this be in generated code? Why can't we copy the output into `types.go` and ditch the need for generating/the csv file?

dyc3:
It avoids having to write out every single one of the cases. I think it's easier.

adapap (approved):
What I meant was to copy the code that was generated and drop it from regenerating every time the same way Gender, Types, and Effectiveness are hard-coded. I guess this way allows us to change the representation easily if needed though.

Issue #85, opened 2021-02-18
Add elemental type matchups to damage calculation


dyc3:

dyc3:
Unassigning: inactive

Issue #204, opened 2021-03-06
Some items don't have FlagUsableInBattle when they should.


dyc3:
## Current Behavior
Some items that are usable in battle don't have `FlagUsableInBattle` or `FlagConsumable` set when they should.

## Expected Behavior
All items that are usable in battle should have this flag.

PR #211, opened 2021-03-08
Held Items

### LINKED ISSUE #281: Items still missing consumable flag
Not sure if this is the only case, but Focus Sash is missing a `FlagConsumable` which is causing errors in my testing of #211 

Related: #204

adapap:
See [held items section](https://veekun.com/dex/items/misc).

All items below implemented, checklist is for tests implemented:
- [x] Black Sludge
- [x] Damp Rock
- [x] Destiny Knot
- [x] Focus Sash
- [x] Heat Rock
- [x] Icy Rock
- [x] King's Rock
- [x] Leftovers
- [x] Life Orb
- [x] Mental Herb
- [x] Muscle Band
- [x] Razor Claw
- [x] Scope Lens
- [x] Shell Bell
- [x] Smooth Rock
- [x] White Herb
- [ ] Wide Lens
- [x] Wise Glasses

## Moved
#243 #244 #245 #246 #247 #248 #249 #250 #251 #252 #253 #255 

This PR also helps me figure out what features need to still be implemented in the "core" of the library.
## Blockers
~#85 (Effectiveness in damage formula) - Expert Belt~

closes #281 (changed for this PR)


dyc3:
It might be a good idea to split this up into multiple PRs, maybe multiple issues. At least for the ones that are blocked by other issues.

adapap:
Added issues for every item that cannot be immediately implemented. No tests are included yet, just want a review on functionality.

battle.go
```diff
@@ -154,6 +155,9 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
 		ctxA := turns[i].Context
 		ctxB := turns[j].Context
 		if reflect.TypeOf(turnA) == reflect.TypeOf(turnB) {
+			if ctxA.Pokemon.HeldItem == ItemQuickClaw {
+				return b.rng.Roll(3, 16) // 3/16 chance to move first
```
dyc3, line 159: This won't work correctly because the RNG is being rolled multiple times.

dyc3 (changes requested):
Just one thing needs to be fixed

battle.go
adapap: Going to move this into a separate issue

move.go
```diff
@@ -59,6 +59,20 @@ type Move struct {
...
+	MinHits       int
+	MaxHits       int
+	MinTurns      int
+	MaxTurns      int
+	Drain         int
+	Healing       int
+	CritRate      int
+	AilmentChance int
+	FlinchChance  int
+	StatChance    int
```
dyc3, line 75: These should probably just go in the `Move` struct (or `MoveData` in PR #256 )

transactions.go
```diff
@@ -192,15 +247,14 @@ func (t InflictStatusTransaction) Mutate(b *Battle) {
 }
 
 type CureStatusTransaction struct {
-	Target       target
+	Target       *Pokemon
```
dyc3, line 250: Leave this as `target`

dyc3 (changes requested):
Oof owie my test coverage. More tests need to be added.

dyc3:
I added a couple of tests. Looks like the Focus Sash doesn't have the consumable flag set, as a result of #204. We should probably fix that before merging this in.

transactions.go
```diff
@@ -150,6 +158,53 @@ func (t ItemTransaction) Mutate(b *Battle) {
...
+			// Need target, change ItemTransaction.Target from *Pokemon to target?
+			// b.QueueTransaction(DamageTransaction{
+			// 	Target: t,
+			// 	Damage: t.Target.MaxHP() / 8,
+			// })
+		}
+	case ItemLeftovers:
+		b.QueueTransaction(HealTransaction{
+			Target: t.Target,
+			Amount: t.Target.MaxHP() / 16,
```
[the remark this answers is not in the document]
adapap: A lot of cases exist where something happens as a fraction of Max HP, with a minimum of 1. This is probably a separate issue that extends beyond just this item.

adapap:
Added all the tests that I could reasonably add. Some blockers/challenges include:

- Some items, as mentioned in #281, are not "consumable" so these tests are failing
- We don't have a way to test for effective accuracy in a battle so there is no way to test "Wide Lens" at the moment -> it would probably need accuracy calculation to be moved outside of damage calculation
- Had a few bugs with transaction utils with Gomega so I attempted to fix them. Also wanted to make the diffing look nicer so made a few QoL changes.

calc.go
```diff
@@ -1,22 +1,25 @@
 package pokemonbattlelib
 
-func calcMoveDamage(weather Weather, user, receiver *Pokemon, move *Move) (damage uint) {
-	weatherMod := 1.0
-	if rain, sun := weather == WeatherRain, weather == WeatherHarshSunlight; (rain && move.Type() == TypeWater) || (sun && move.Type() == TypeFire) {
-		weatherMod = 1.5
+func calcMoveDamage(b *Battle, user, receiver *Pokemon, move *Move) (damage float64) {
```
dyc3, line 3: Don't change this function signature, it intentionally doesn't take a `*Battle` and returns a `uint`

calc.go
```diff
@@ -1,22 +1,25 @@
...
+	weather := 1.0
+	if rain, sun := b.Weather == WeatherRain, b.Weather == WeatherHarshSunlight; (rain && move.Type() == TypeWater) || (sun && move.Type() == TypeFire) {
+		weather = 1.5
 	} else if (rain && move.Type() == TypeFire) || (sun && move.Type() == TypeWater) {
-		weatherMod = 0.5
+		weather = 0.5
+	}
+	crit := 1.0
+	if b.rng.Roll(1, user.CritChance()) {
+		crit = 2.0
```
dyc3, line 12: This was intentionally moved outside of this function because it's random. Please put it back where it was.

data/gen.go
```diff
@@ -44,6 +44,23 @@ type data_pokemon struct {
...
+	maxHits       int
+	minTurns      int
+	maxTurns      int
+	drain         int
+	healing       int
+	critRate      int
+	ailmentChance int
+	flinchChance  int
+	statChance    int
+}
```
dyc3, line 58: This is just more move data, it should just go in `data_move`.

data/gen.go
```diff
@@ -555,10 +573,36 @@ func main() {
...
+				continue
+			}
+			moves[i].Meta = data_move_meta{minHits, maxHits, minTurns, maxTurns, drain, healing, critRate, ailmentChance, flinchChance, statChance}
+		}
+	}
 	output += "var AllMoves = []MoveData{\n"
 	for _, p := range moves {
-		output += fmt.Sprintf("\t{Name: %q, Type: %d, Category: %s,"+
-			" Targets: %d, Priority: %d, Power: %d, Accuracy: %d},\n", p.Name, p.Type, p.DamageClass, p.Targets, p.Priority, p.Power, p.Accuracy)
+		output += fmt.Sprintf("\t{Name: %q, Type: %d, Category: %s, Targets: %d, Priority: %d, Power: %d, Accuracy: %d, metadata: %s},\n",
```
dyc3, line 604: There is no need to have a separate struct `MoveMeta` inside of `MoveData`. Just make all the fields of `MoveMeta` fields of `MoveData` instead.

Also, the format string `%#v` outputs the struct as go code.

calc.go
```diff
@@ -27,24 +30,37 @@ func calcMoveDamage(weather Weather, user, receiver *Pokemon, move *Move) (damag
...
+	case ItemLifeOrb:
+		modifier *= 1.30
+	case ItemMuscleBand:
+		if move.Category() == MoveCategoryPhysical {
+			modifier *= 1.10
+		}
+	case ItemWiseGlasses:
+		if move.Category() == MoveCategorySpecial {
+			modifier *= 1.10
+		}
```
dyc3, line 62: Would prefer these to be integer operations

dyc3 (changes requested):

util_test.go
```diff
@@ -310,6 +319,17 @@ func (g *TestRNG) Roll(x, y int) bool {
...
+
+// Always rolls random effects
+func AlwaysRNG() *TestRNG {
+	return &TestRNG{rolls: []bool{true}}
+}
+
+// Always hit, never crit
+func SimpleRNG() *TestRNG {
+	return &TestRNG{rolls: []bool{true, false}}
+}
```
dyc3, line 335: Could you move this change to a separate PR?

calc.go
adapap: How do you do integer operations? Best way i've come up with is:
```golang
x *= 1.35
// in integer operation
x = (x * 135) / 100
```

Can we add some utility functions to perform floating/percentage modifiers on `uint` values?

dyc3: Yeah, that's basically it. I don't think it would be very beneficial to add utility functions just to do a multiply and divide

battle.go
```diff
@@ -126,8 +126,14 @@ func (b *Battle) Start() error {
...
 		if v, ok := t.Pokemon.metadata[MetaSleepTime]; ok && v.(int) == 0 && t.Pokemon.StatusEffects.check(StatusSleep) {
+			pkmn := b.getPokemon(t)
 			b.QueueTransaction(CureStatusTransaction{
-				Target:       t,
+				Target: target{
+					party:     0,
+					partySlot: 0,
+					Team:      0,
+					Pokemon:   *pkmn,
+				},
```
dyc3, line 136: Change this back, it breaks sleeping for pokemon that are not in party 0 slot 0

battle_test.go
```diff
@@ -1019,6 +1019,62 @@ var _ = Describe("Fainting", func() {
...
+				User: b.getPokemonInBattle(1, 0),
+				Target: target{
+					Pokemon:   *b.getPokemonInBattle(0, 0),
+					party:     0,
+					partySlot: 0,
+					Team:      0,
+				},
+			}))
+			Expect(b.parties[0].activePokemon[0].CurrentHP).To(BeEquivalentTo(1))
+		})
```
dyc3, line 1052: Needs to assert that there was no `FaintTransaction`

move.go
dyc3: These should go in `MoveData` directly, and the `MoveMeta` type should not be necessary.

dyc3 (changes requested):
Let's try to avoid big PRs like this in the future.

transactions.go
```diff
@@ -161,9 +210,18 @@ type PPTransaction struct {
...
+		n := uint8(t.Amount * -1)
+		if t.Move.CurrentPP < n {
+			n = t.Move.CurrentPP
+		}
+		t.Move.CurrentPP -= n
+	} else {
+		t.Move.CurrentPP += uint8(t.Amount)
+		if t.Move.CurrentPP > t.Move.MaxPP {
+			t.Move.CurrentPP = t.Move.MaxPP
+		}
```
dyc3, line 224:
```suggestion
	t.Move.CurrentPP += uint8(t.Amount)
	if t.Move.CurrentPP >= t.Move.MaxPP {
		if t.Amount > 0 {
			t.Move.CurrentPP = t.Move.MaxPP
		} else {
			t.Move.CurrentPP = 0
		}
	}
```

dyc3 (changes requested):

dyc3 (approved):
I think we're good to go on this. All we need to do is fix the merge conflicts.

Definitely make these smaller in the future though, this took way too long to merge in

PR #218, opened 2021-03-08
Better test messages

### LINKED ISSUE #186: Simplify `HaveTransaction` parameters?
While most of the time it is helpful to explicitly state all of the fields of the transactions expected, sometimes doing so leads to undecipherable error messages or extra fields not needed for the specific test. Examples are including `party` and `partySlot` which default to 0 already, or the error message from the transactions not matching.

It might make more sense to have something like this:
```
Expect(transactions).To(HaveTransaction(DamageTransaction))
Expect(poke.CurrentHP).To(BeEquivalentTo(15))
```

adapap:
- Removed HaveLen from some tests as this is implementation specific, and we already have `HaveTransaction` and `HaveTransactionsInOrder` for those tests.

Currently, the test message you get if you have an incorrect field in the transaction is as such:
```go
Expect(t).To(HaveTransaction(DamageTransaction{
	Target: target{
		Pokemon:   *pkmn1,
		party:     0,
		partySlot: 0,
		Team:      0,
	},
	Damage:       11, // supposed to be 12
	StatusEffect: StatusBadlyPoison,
}))
```

>Expected the sequence of transactions to include: pokemonbattlelib.DamageTransaction. 3 of the same type were found, but none of them matched.

This has been changed to show a "diff" of the fields that do not match. I added a string representation of target since that gets used frequently. The result looks like this for the same transaction:
> Closest pokemonbattlelib.DamageTransaction (from 3 total) has 2 fields that do not match:
    - Target
    Party 0 (Slot 0) | Team 0 | Pokemon:
    Bulbasaur   Lv100
    HP: 112/200
    - Damage
    12

(TODO: support compare transaction logic for checking pointers and stuff)

Closes #186


util_test.go
```diff
@@ -151,11 +174,13 @@ func (matcher *singleTransactionMatcher) FailureMessage(actual interface{}) (mes
...
-			return fmt.Sprintf("Expected the sequence of transactions to include: %T. %d of the same type were found, but none of them matched.",
-				matcher.expected,
-				count,
-			)
+			expected := ""
+			closest := getClosestTransaction(transactions, matcher.expected)
+			for name, val := range closest {
+				expected += fmt.Sprintf("- %s\n%s\n", name, val)
+			}
+			return fmt.Sprintf("Closest %T (from %d total) has %d fields that do not match:\n%s",
```
dyc3, line 182: would prefer "x/y fields that do not match" or "x out of y fields that do not match"

dyc3 (changes requested):
I tried changing the test "should inflict badly poisoned damage", changing the damage checked to be different, and it gave me the output:
```
Closest pokemonbattlelib.DamageTransaction (from 3 total) has 2 fields that do not match:
    - Target
    Party 0 (Slot 0) | Team 0 | Pokemon:
    Bulbasaur   Lv100
    HP: 112/200
    
    - Damage
    12
```
It should have only given one field.

I'd also like to see what the values it expected were, eg
```
Damage: got 14, expected 12
```

adapap:
> It should have only given one field.

Yes, I wasn't sure how to best compare non-trivial structs involving pointers or nested structs. I think I will reuse the logic from `compareTransaction`.
> I'd also like to see what the values it expected were

Also wasn't sure the best way to represent this for larger structs without some complicated recursive strategy.

adapap:
Alright this looks sweet! Here is a sample error message:
```
Closest pokemonbattlelib.DamageTransaction (3 total) has 2/5 fields that don't match:
    - Target
    Expected: Party 1 (Slot 0) | Team 0 | Pokemon:
    Articuno    Lv1
    HP: 6/12

    Received: Party 0 (Slot 0) | Team 0 | Pokemon:
    Articuno    Lv1
    HP: 12/12

    - Damage
    Expected: 7
    Received: 6
```

util_test.go
```diff
@@ -108,6 +128,28 @@ func findCountTransactionIdxWithMatchingType(transactions []Transaction, a Trans
...
+	for _, t := range check {
+		result := transactionDiff(want, t)
+		// v := reflect.ValueOf(t)
+		// for j := 0; j < v.NumField(); j += 1 {
+		// 	a := v.Field(j).Interface()
+		// 	b := expected.Field(j).Interface()
+		// 	if !reflect.DeepEqual(a, b) {
+		// 		result[v.Type().Field(j).Name] = fmt.Sprintf("%v", a)
+		// 	}
+		// }
```
dyc3, line 144: remove commented code

dyc3 (changes requested):
Looking good!

dyc3 (approved):

PR #223, opened 2021-03-09
Finish implementing Sleep

### LINKED ISSUE #222: Finish implementing Sleep status condition
See: https://github.com/StevensSEC/pokemonbattlelib/wiki/Requirements#non-volatile-status-conditions

TODO:
- [x] Wake pokemon up after random number of turns between 1 and 5 inclusive

Moves:
- [x] [Snore](https://bulbapedia.bulbagarden.net/wiki/Snore_(move)) can be used while sleeping, and turns count towards waking the pokemon
- [x] [Sleep Talk](https://bulbapedia.bulbagarden.net/wiki/Sleep_Talk_(move)) can be used while sleeping, and turns count towards waking the pokemon

dyc3:
- allow sleeping pokemon to wake up
- allow Snore and Sleep Talk while sleeping

closes #222

### DIFF
--- a/battle.go
+++ b/battle.go
@@ -108,7 +108,22 @@ func (b *Battle) Start() error {
 
 // Handles all pre-turn logic
 func (b *Battle) preRound() {
-	// TODO
+	for a, party := range b.parties {
+		for ap, pkmn := range party.activePokemon {
+			t := target{
+				party:     a,
+				partySlot: ap,
+				Pokemon:   *pkmn,
+				Team:      party.team,
+			}
+			if v, ok := pkmn.metadata[MetaSleepTime]; ok && v.(int) == 0 && pkmn.StatusEffects.check(StatusSleep) {
+				b.QueueTransaction(CureStatusTransaction{
+					Target:       t,
+					StatusEffect: StatusSleep,
+				})
+			}
+		}
+	}
 }
 
 // Simulates a single round of the battle. Returns processed transactions for this turn and indicates whether the battle has ended.
@@ -117,6 +132,7 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
 		log.Panic("battle is not currently in progress")
 	}
 	b.preRound()
+	b.ProcessQueue()
 	// Collects all turn info from each active Pokemon
 	turns := make([]TurnContext, 0)
 	for i, party := range b.parties {
--- a/battle.go
+++ b/battle.go
@@ -189,6 +189,7 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
 		switch t := turn.Turn.(type) {
 		case FightTurn:
 			user := turn.Context.Pokemon
+			move := user.Moves[t.Move]
 			// pre-move checks
 			if user.StatusEffects.check(StatusFreeze) || user.StatusEffects.check(StatusParalyze) {
 				immobilize := false
@@ -206,7 +207,7 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
 					})
 					continue // forfeit turn
 				}
-			} else if user.StatusEffects.check(StatusSleep) {
+			} else if user.StatusEffects.check(StatusSleep) && move.ID != MoveSnore && move.ID != MoveSleepTalk {
 				b.QueueTransaction(ImmobilizeTransaction{
 					Target: target{
 						Pokemon: user,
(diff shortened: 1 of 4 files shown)


battle.go
```diff
@@ -108,7 +108,22 @@ func (b *Battle) Start() error {
...
+				Team:      party.team,
+			}
+			if v, ok := pkmn.metadata[MetaSleepTime]; ok && v.(int) == 0 && pkmn.StatusEffects.check(StatusSleep) {
+				b.QueueTransaction(CureStatusTransaction{
+					Target:       t,
+					StatusEffect: StatusSleep,
+				})
+			}
+		}
+	}
```
adapap, line 126: This particular case can be simplified to `for _, t := range b.GetTargets()` but I understand if we want to keep this for future cases.

battle.go
```diff
@@ -117,6 +132,7 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
 		log.Panic("battle is not currently in progress")
 	}
 	b.preRound()
+	b.ProcessQueue()
```
adapap, line 135: Move to `b.preRound()`?

battle.go
```diff
@@ -190,7 +207,7 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
 					})
 					continue // forfeit turn
 				}
-			} else if user.StatusEffects.check(StatusSleep) {
+			} else if user.StatusEffects.check(StatusSleep) && move.ID != MoveSnore && move.ID != MoveSleepTalk {
 				b.QueueTransaction(ImmobilizeTransaction{
 					Target: target{
 						Pokemon: user,
```
adapap, line 213: (meta) Since this is just gonna keep growing, it might be good to split up SimulateRound some more, like `b.getTurns() []TurnContext`, `b.sortTurns(turns *[]TurnContext)`. Probably don't need a separate function to process turns, we can keep that in `SimulateRound`.

battle_test.go
```diff
@@ -1067,26 +1067,112 @@ var _ = Describe("Status Conditions", func() {
...
+				pkmn1 := GeneratePokemon(PkmnBulbasaur, WithLevel(8), WithMoves(GetMove(MovePound)))
+				p1 = NewOccupiedParty(&a1, 0, pkmn1)
+				pkmn2 := GeneratePokemon(PkmnCharmander, WithLevel(4), WithMoves(GetMove(MovePound)))
+				p2 = NewOccupiedParty(&a2, 1, pkmn2)
+				b.AddParty(p1, p2)
+				Expect(b.Start()).To(Succeed())
+				b.QueueTransaction(InflictStatusTransaction{
+					Target:       pkmn1,
+					StatusEffect: StatusSleep,
+				})
```
adapap, line 1081: None of the other tests queue transactions directly, so I think this should implement any of the sleep causing mechanisms (`MoveSleepPowder` for example).

battle_test.go
```diff
@@ -1067,26 +1067,112 @@ var _ = Describe("Status Conditions", func() {
...
+							party:     0,
+							partySlot: 0,
+							Team:      0,
+						},
+						StatusEffect: StatusSleep,
+					},
+				))
+			})
+
+			It("should allow sleeping Pokemon to use Sleep Talk", func() {
```
adapap, line 1151: Can probably merge/table with previous test.

transactions.go
```diff
@@ -145,6 +148,9 @@ type CureStatusTransaction struct {
 func (t CureStatusTransaction) Mutate(b *Battle) {
 	receiver := b.getPokemon(t.Target.party, t.Target.partySlot)
 	receiver.StatusEffects.clear(t.StatusEffect)
+	if t.StatusEffect.check(StatusSleep) {
+		delete(receiver.metadata, MetaSleepTime)
```
adapap, line 152: (nit) I had this initialize and reset the meta field to 0 to avoid checking for existence in `b.preRound`.

transactions.go
```diff
@@ -223,7 +229,10 @@ type ImmobilizeTransaction struct {
 }
 
 func (t ImmobilizeTransaction) Mutate(b *Battle) {
-	// currently a no-op.
+	receiver := b.getPokemon(t.Target.party, t.Target.partySlot)
+	if t.StatusEffect.check(StatusSleep) {
+		receiver.metadata[MetaSleepTime] = receiver.metadata[MetaSleepTime].(int) - 1
+	}
```
adapap, line 235: I don't think this is covered by tests.

adapap (commented):
Room to simplify, but just a few thoughts:

battle_test.go
dyc3: I did this to keep the test as focused on what it's testing as possible.

adapap (approved):

PR #229, opened 2021-03-09
fix: change Pokemon.HeldItem to be an ID (int)

adapap:
This doesn't really add syntactic sugar if you need any other properties of an item, such as the category in this example:

```go
if pkmn.HeldItem != 0 && pkmn.CurrentHP <= pkmn.Stats[StatHP]/4 {
	if i := GetItem(pkmn.HeldItem); i.Category == ItemCategoryInAPinch {
		b.QueueTransaction(ItemTransaction{
			Target: pkmn,
			Item:   i,
		})
	}
}
```

### DIFF
--- a/transactions.go
+++ b/transactions.go
@@ -88,14 +88,14 @@ func (t EVTransaction) Mutate(b *Battle) {
 // A transaction to use and possibly consume an item.
 type ItemTransaction struct {
 	Target *Pokemon
-	Item   *Item
+	Item   Item
 	Move   *Move
 }
 
 func (t ItemTransaction) Mutate(b *Battle) {
 	if t.Item.Flags&FlagConsumable > 0 {
-		if t.Target.HeldItem == t.Item {
-			t.Target.HeldItem = nil
+		if t.Target.HeldItem == t.Item.ID {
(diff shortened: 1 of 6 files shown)


transactions.go
```diff
@@ -88,14 +88,14 @@ func (t EVTransaction) Mutate(b *Battle) {
...
+	Item   Item
 	Move   *Move
 }
 
 func (t ItemTransaction) Mutate(b *Battle) {
 	if t.Item.Flags&FlagConsumable > 0 {
-		if t.Target.HeldItem == t.Item {
-			t.Target.HeldItem = nil
+		if t.Target.HeldItem == t.Item.ID {
+			t.Target.HeldItem = 0
```
dyc3, line 98: This doesn't really hold up, because if an agent uses the same item that the pokemon is holding, the held item would be consumed when it shouldn't.

adapap:
I don't really think this is the solution, period. It might be better to just keep the code using pointers, and add helper functions for syntactic sugar because that's a simple way to check for existence and compare simultaneously.

dyc3:
Let's close this, and I'll take a stab at it

PR #230, opened 2021-03-09
general improvements to http api

dyc3:


http/api.go
```diff
@@ -157,7 +157,10 @@ func HandleBattleSimulate(w http.ResponseWriter, r *http.Request) {
 
 	bytes, err := json.Marshal(results)
 	if err != nil {
-		log.Fatalf("Failed to marshal into JSON: %s", err)
+		log.Printf("Failed to marshal into JSON: %s", err)
```
A-very-Cunning-ham, line 160: is there a reason this isn't a Panicf like the similar JSON error in agent.go?

dyc3: The panic in agent.go is to get it to leave the function and fail. The print here is to let the server respond to the request.

A-very-Cunning-ham: makes sense!

A-very-Cunning-ham (approved):
looking good

PR #231, opened 2021-03-10
item refactor

### LINKED ISSUE #227: Check if Pokemon is holding an item
Syntactic sugar for this pattern:
`if p.HeldItem != nil && p.HeldItem.ID == itemID`

I'm thinking one of these solutions:
`p.GetHeldItem() int` // returns item ID, or -1
`p.HasHeldItem(...id int) bool` // returns true if Pokemon is holding any of the given items

dyc3:
Should close #227


adapap (changes requested):
The interface is nicer to use but here are a couple of concerns:
- The program will crash differently from getting invalid moves/Pokemon by ID without the panic in `GetItem`
- The item handling should probably be handled in `items.go` to limit the length of `transactions.go`. This can be a simple call like `b.UseItem(Item)`, which can perform the same functions as `Mutate`
- I'm unclear on how certain side effects will be handled because I don't think they can be done in `Mutate`: examples are `ItemQuickClaw` (chance for Pokemon to attack first) and `ItemCustapBerry` (forces Pokemon to go first), which both affect priority, as well as items that affect accuracy for damage transactions

dyc3:
> The item handling should probably be handled in items.go to limit the length of transactions.go.

Let's hold off on that until it becomes a bigger problem. We'll be able to come up with a better solution when we have more items implemented.

> I'm unclear on how certain side effects will be handled because I don't think they can be done in Mutate: examples are ItemQuickClaw (chance for Pokemon to attack first) and ItemCustapBerry (forces Pokemon to go first), which both affect priority, as well as items that affect accuracy for damage transactions

We can easily put those checks in the preround, and then set a metadata entry. When the turns get sorted, they can check for the metadata entry.

adapap (approved):

PR #269, opened 2021-03-20
party validation

### LINKED ISSUE #147: Validate Pokemon for battle
Possible duplicate of #34.

Pokemon should be validated before a battle begins. Some of the battle codes assumes that a Pokemon has moves, so we want to avoid nil references if no moves are provided.

See [requirements for Pokemon](https://github.com/StevensSEC/pokemonbattlelib/wiki/Requirements#pokemon).

dyc3:
closes #147


party.go
```diff
@@ -11,12 +12,14 @@ type party struct {
...
 	activePokemon map[int]*Pokemon // Map containing slots and references to active Pokemon on the battlefield
 	team          int              // The team that this party belongs to
+	PokemonRules  PokemonValidationRules
 }
 
 // Maximum number of Pokemon in a party
 const MaxPartySize = 6
 
 var PartyIndexError = errors.New("invalid index for party")
+var ErrorPartyFull = fmt.Errorf("party size cannot exceed max of %d Pokemon\n", MaxPartySize)
```
adapap, line 22: Change either `PartyIndexError` or `ErrorPartyFull` to match the other

battle_test.go
```diff
@@ -6,6 +6,8 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+var defaultMoveOpt = WithMoves(GetMove(MovePound))
```
adapap, line 9: Do we want to have some kind of no-op move as a default in `GeneratePokemon` instead? I feel like most cases for using pound are to create a valid Pokemon with 1 move that serves no purpose.

Edit: I see the new issue for this now.

adapap (changes requested):
Interesting approach with flags, might be overkill but I guess it's customizable.

adapap (approved):

PR #284, opened 2021-03-23
Add type matchups to damage calculation

### LINKED ISSUE #85: Add elemental type matchups to damage calculation

dyc3:
This PR also refactors how the `Effectiveness` type works such that no casts to `float64` are necessary.

closes #85

### DIFF
--- a/battle_test.go
+++ b/battle_test.go
@@ -266,7 +266,7 @@ var _ = Describe("One round of battle", func() {
 					Team:      1,
 				},
 				Move:   GetMove(MoveFireFang),
-				Damage: 9,
+				Damage: 8,
 			}))
 		})
 
--- a/battle_test.go
+++ b/battle_test.go
@@ -206,68 +206,123 @@ var _ = Describe("One round of battle", func() {
 			Expect(bidoof.CurrentHP).To(BeEquivalentTo(93))
 		})
 
-		It("should account for supereffective type matchups", func() {
-			a1 := newRcAgent()
-			a2 := newRcAgent()
-			_a1 := Agent(a1)
-			_a2 := Agent(a2)
-			pkmn1 := GeneratePokemon(
-				PkmnMightyena,
-				WithIVs([6]uint8{31, 0, 31, 0, 31, 31}),
-				WithMoves(
-					GetMove(MoveFireFang),
-					GetMove(MoveTackle),
-				),
+		Context("Type Matchups", func() {
+			var (
+				a1  rcAgent
+				a2  rcAgent
+				_a1 Agent
+				_a2 Agent
+				b   *Battle
 			)
-			party1 = NewOccupiedParty(&_a1, 0, pkmn1)
-			pkmn2 := GeneratePokemon(
-				PkmnTurtwig,
-				WithMoves(GetMove(MoveTackle)),
-				WithIVs([6]uint8{31, 31, 31, 31, 31, 0}),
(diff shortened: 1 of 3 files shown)


battle_test.go
```diff
@@ -206,6 +206,70 @@ var _ = Describe("One round of battle", func() {
...
+				WithIVs([6]uint8{31, 31, 31, 31, 31, 0}),
+			)
+			party2 = NewOccupiedParty(&_a2, 1, pkmn2)
+			b := NewBattle()
+			b.AddParty(party1, party2)
+			b.rng = &SimpleRNG
+			Expect(b.Start()).To(Succeed())
+
+			// TODO: test the difference in damage between the transactions rather than the exact values of the transactions
+			// TODO: make it so that target doesn't need to include `Pokemon` or `Team`
```
adapap: Maybe we want a `b.getTarget(party, slot int) target`?

battle_test.go
```diff
@@ -206,6 +206,125 @@ var _ = Describe("One round of battle", func() {
...
+				_a1 Agent
+				_a2 Agent
+				b   *Battle
+			)
+
+			BeforeEach(func() {
+				a1 = newRcAgent()
+				a2 = newRcAgent()
+				_a1 = Agent(a1)
+				_a2 = Agent(a2)
```
adapap, line 222: Doesn't need to be repeated, just define it as a variable above.

battle_test.go
```diff
@@ -206,6 +206,125 @@ var _ = Describe("One round of battle", func() {
...
+					PkmnTurtwig,
+					WithMoves(GetMove(MoveTackle)),
+					WithIVs([6]uint8{31, 31, 31, 31, 31, 0}),
+				)
+				party2 = NewOccupiedParty(&_a2, 1, pkmn2)
+				b.AddParty(party1, party2)
+				Expect(b.Start()).To(Succeed())
+
+				// TODO: test the difference in damage between the transactions rather than the exact values of the transactions
+				// TODO: make it so that target doesn't need to include `Pokemon` or `Team`
```
adapap, line 247: Maybe we want a `b.getTarget(party, slot int) target`?

calc.go
```diff
@@ -46,5 +46,11 @@ func calcMoveDamage(weather Weather, user, receiver *Pokemon, move *Move) (damag
 		}
 	}
 	damage = uint((((levelEffect * movePower * attack / defense) / 50) + 2) * modifier)
+	elementalEffect := GetElementalEffect(move.Type(), receiver.EffectiveType())
```
adapap, line 49: I'm not sure exactly how the order works in the damage calculation, but adding this multiplier after truncating to `uint` seems inconsistent. I feel like `modifier`, which includes all values such as STAB and type effectiveness should be consolidated before applying it to the damage formula.

adapap (changes requested):
OK - hope this doesn't destroy too many tests :-{

calc.go
dyc3: I want to refactor the damage calculation to remove all the casts to `float64`, because they cause tests to be extremely flakey.

Since all the modifier multipliers come after the initial damage calculation, doing a bitshift like this is functionally equivalent. This implementation also feels more clean because `NoEffect` matchups don't require a special case.

If anything, it's more consistent because it's not a working with a float64.

adapap (approved):

PR #291, opened 2021-03-26
Better Test Utilities

### LINKED ISSUE #289: Cannot use `defaultMoveOpt` in tests that check for moves
When using `defaultMoveOpt`, there is no way to get the pointer to match the `Move` field of a transaction.

adapap:
This PR adds more descriptive messages to failing tests.

## Example 1
A missing/incorrect field in ordered transactions is difficult to deduce from the error message.

For the following code:
```golang
Expect(t).To(HaveTransactionsInOrder(
	HealTransaction{
		Target: charmander,
		Amount: 0,
	},
	DamageTransaction{
		User: bulbasaur,
		Target: target{
			Pokemon:   *charmander,
			party:     1,
			partySlot: 0,
			Team:      1,
		},
		Move:   GetMove(MoveTackle), // Should be MovePound
		Damage: 3,
	},
))
```

The current error is as follows:
```
Expected the sequence of transactions to have these transactions in this order:
    0: pokemonbattlelib.HealTransaction: {Target:Charmander Amount:0}
    1: pokemonbattlelib.DamageTransaction: {User:Bulbasaur Target:Party 1 (Slot 0) | Team 1 | Pokemon:
    Charmander Move:Fire Blast Damage:3 StatusEffect:}
```

This is my attempt at a solution:
```
Expected the sequence of transactions to have these transactions in this order:
    1. pokemonbattlelib.DamageTransaction
    2. pokemonbattlelib.DamageTransaction

    Received the following transactions:
    1. pokemonbattlelib.DamageTransaction
    2. pokemonbattlelib.DamageTransaction

    The closest transaction that failed to match is shown below:
    pokemonbattlelib.DamageTransaction has 1/5 fields that do not match:
    - Move
    Expected: Tackle
    Received: Pound
```
## Example 2
When matching single transactions, it is sometimes tricky to see which fields are affected when two transactions are not equivalent.
The following error is not the best example, but I've had cases where it's really difficult to find the field that is wrong, especially when working with arrays.
```Expected:
        pokemonbattlelib.DamageTransaction: {User:Bulbasaur Target:Party 0 (Slot 0) | Team 1 | Pokemon:
    Charmander Move:Pound Damage:3 StatusEffect:}

    Instead, got:
        pokemonbattlelib.DamageTransaction: {User:Bulbasaur Target:Party 1 (Slot 0) | Team 1 | Pokemon:
    Charmander Move:Pound Damage:3 StatusEffect:}
```

Becomes this:
```
pokemonbattlelib.DamageTransaction has 1/5 fields that do not match:
    - Target
    Expected: Party 0 (Slot 0) | Team 1 | Pokemon:
    Charmander
    Received: Party 1 (Slot 0) | Team 1 | Pokemon:
    Charmander
```

## Example 3
We should be able to see which transactions occurred instead of knowing that one is missing.

Currently, we get the following message:
```
Expected the sequence of transactions to include: pokemonbattlelib.InflictStatusTransaction, but none of the same type were found in 2 transactions.
```

Now, we get this message:
```
Expected the sequence of transactions to include pokemonbattlelib.InflictStatusTransaction, but received:
    - pokemonbattlelib.DamageTransaction
    - pokemonbattlelib.HealTransaction
```

Additionally, made `Move` check a strict match similar to how Pokemon are compared in the matchers.
Closes #289

### DIFF
--- a/util_test.go
+++ b/util_test.go
@@ -180,6 +180,17 @@ func getClosestTransaction(check []Transaction, want Transaction) map[string]dif
 	return best
 }
 
+// Creates a text representation of the diff result of transactions
+func getDiffText(check []Transaction, want Transaction) string {
+	diffText := ""
+	closest := getClosestTransaction(check, want)
+	total := reflect.ValueOf(want).NumField()
+	for name, val := range closest {
+		diffText += fmt.Sprintf("- %s\nExpected: %v\nReceived: %v\n", name, val.expected, val.got)
+	}
+	return fmt.Sprintf("%T has %d/%d fields that do not match:\n%s", want, len(closest), total, diffText)
+}
+
 // Given a sequence of transactions, match if a given transaction is present in the sequence.
 type singleTransactionMatcher struct {
 	expected Transaction
@@ -223,14 +234,8 @@ func (matcher *singleTransactionMatcher) FailureMessage(actual interface{}) (mes
 				transactions[first],
 			)
 		} else {
-			diffText := ""
-			closest := getClosestTransaction(transactions, matcher.expected)
-			total := reflect.ValueOf(matcher.expected).NumField()
-			for name, val := range closest {
-				diffText += fmt.Sprintf("- %s\nExpected: %v\nReceived: %v\n", name, val.expected, val.got)
-			}
-			return fmt.Sprintf("Closest %T (%d total) has %d/%d fields that don't match:\n%s",
-				matcher.expected, count, len(closest), total, diffText)
+			diffText := getDiffText(transactions, matcher.expected)
+			return fmt.Sprintf("The closest of %d total %s", count, diffText)
 		}
 	default:
 		return fmt.Sprintf("Actual's type is %T, when it should be []Transaction", actual)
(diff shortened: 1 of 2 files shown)


util_test.go
```diff
@@ -100,6 +100,13 @@ func transactionDiff(expected, got Transaction) map[string]diff {
 					got:      rfB.Interface(),
 				}
 			}
+		} else if rfA.Type() == reflect.TypeOf(&Move{}) {
+			if !rfB.IsNil() && !reflect.DeepEqual(rfA.Interface(), rfB.Interface()) {
+				result[typeField.Name] = diff{
+					expected: rfA.Interface(),
+					got:      rfB.Interface(),
+				}
+			}
```
dyc3, line 109: Can we just compare the move id instead? Because really all we want to do when comparing transactions is check to see if the correct move was used. If we want to check PP amounts, we're probably going to check the pokemon itself, not the transaction.

dyc3 (changes requested):

util_test.go
dyc3: I'm just gonna go ahead and make this change. I need the fix in this PR.

dyc3 (approved):

PR #295, opened 2021-03-29
Refactor target to have ground truth pointer

### LINKED ISSUE #294: Agents are not given copies of moves.
## Current Behavior
Agents are not given copies of moves. Agents should not receive ground truth pointers to the Pokemon's moves.

**To Reproduce**
```go
It("should never receive a ground truth reference to pokemon in the battle", func() {
		b := NewBattle()
		a1 := newRcAgent()
		_a1 := Agent(a1)
		a2 := Agent(funkyAgent(func(ctx *BattleContext) Turn {
			real := b.getPokemon(ctx.Opponents[0])
			Expect(ctx.Opponents[0].Pokemon).ToNot(BeIdenticalTo(real))
			Expect(ctx.Opponents[0].Pokemon.Moves[0]).ToNot(BeIdenticalTo(real.Moves[0]))
			return FightTurn{Move: 0, Target: ctx.Opponents[0]}
		}))
		party1 := NewOccupiedParty(&_a1, 0, GeneratePokemon(PkmnCharmander, defaultMoveOpt))
		party2 := NewOccupiedParty(&a2, 1, GeneratePokemon(PkmnSquirtle, defaultMoveOpt))
		b.AddParty(party1, party2)
		Expect(b.Start()).To(Succeed())
		a1 <- FightTurn{
			Move: 0,
			Target: target{
				party:     1,
				partySlot: 0,
			},
		}
		b.SimulateRound()
	})
```

## Expected Behavior
Expected `Moves` to be copied when given to an Agent.

**Additional context**

dyc3:
- add agent test
- refactor target to contain a pointer to the ground truth pokemon instead of a copy

This PR is an effort to consolidate how pokemon are accessed into one easy mechanism. An added bonus is a potentially reduced lifetime for values copied when creating battle contexts for Agents, which would mean less memory usage over general usage.

Now the code more clearly defines where pokemon are copied, and where pokemon are passed by reference.

Another bonus, it closes #294

### DIFF
--- a/battle.go
+++ b/battle.go
@@ -138,20 +138,20 @@ func (b *Battle) sortTurns(turns *[]TurnContext) {
 	sort.SliceStable(*turns, func(i, j int) bool {
 		turnA := (*turns)[i].Turn
 		turnB := (*turns)[j].Turn
-		ctxA := (*turns)[i].Context
-		ctxB := (*turns)[j].Context
+		pkmnA := (*turns)[i].User.Pokemon
+		pkmnB := (*turns)[j].User.Pokemon
 		if reflect.TypeOf(turnA) == reflect.TypeOf(turnB) {
 			switch turnA.(type) {
 			case FightTurn:
 				ftA := turnA.(FightTurn)
 				ftB := turnB.(FightTurn)
-				mvA := ctxA.Pokemon.Moves[ftA.Move]
-				mvB := ctxB.Pokemon.Moves[ftB.Move]
+				mvA := pkmnA.Moves[ftA.Move]
+				mvB := pkmnB.Moves[ftB.Move]
 				if mvA.Priority() != mvB.Priority() {
 					return mvA.Priority() > mvB.Priority()
 				}
 				// speedy pokemon should go first
-				return ctxA.Pokemon.Speed() > ctxB.Pokemon.Speed()
+				return pkmnA.Speed() > pkmnB.Speed()
 			}
 		} else {
 			// make higher priority turns go first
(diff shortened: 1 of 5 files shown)


battle.go
```diff
@@ -481,28 +484,53 @@ type BattleContext struct {
...
 
-// Gets all the active Pokemon (targets) in the battle
+// Gets a deep copy of all the active Pokemon (targets) in the battle
 func (b *Battle) GetTargets() []target {
 	targets := make([]target, 0)
 	for partyID, party := range b.parties {
 		for slot, active := range party.activePokemon {
+			var pkmn Pokemon
+			bytes, _ := json.Marshal(active)
+			json.Unmarshal(bytes, &pkmn)
```
adapap, line 494: This seems like a gimmicky way of making a deep copy. Why can't we use a golang lib function?

battle_test.go
```diff
@@ -296,13 +318,13 @@ var _ = Describe("One round of battle", func() {
 				Expect(b.Start()).To(Succeed())
 
 				// TODO: make it so that target doesn't need to include `Pokemon` or `Team`
```
adapap, line 320: Let's make an issue for this if not done already

adapap (approved):

battle.go
dyc3: The sort answer is no, [Go does not have a standard library function to deep copy an object](https://stackoverflow.com/questions/56355212/deep-copying-data-structures-in-golang), for some reason. I saw somebody else's benchmarks, and this was the fastest.

PR #302, opened 2021-04-03
add pokemon to battle results

dyc3:


adapap (approved):
I guess you don't need a reference to the party for the results?

dyc3:
This was made before the other PR that added the `Party` type. I suppose it would be better to pass the party instead.

PR #303, opened 2021-04-03
add Party type, and refactor most tests for consistency

dyc3:
This keeps but renames the old party type to `battleParty`, which performs
the same function as `party` did before. This will make it easier for users
to reuse a `Party` outside of a battle.

While I was at it, I also refactored most of the tests to be a little more
consistent, using new `NewSingleBattle` and `New1v1Battle` test utilities.

Related #237

### DIFF
--- a/util_test.go
+++ b/util_test.go
@@ -374,3 +374,40 @@ func AlwaysRNG() *TestRNG {
 func SimpleRNG() *TestRNG {
 	return &TestRNG{rolls: []bool{true, false}}
 }
+
+// Create a single Battle. A single battle is a battle where only 1 pokemon is sent out at a time, and there can only be 2 parties.
+// This function should only be used in tests.
+func NewSingleBattle(p1 *Party, a1 *Agent, p2 *Party, a2 *Agent) *Battle {
+	b := NewBattle()
+	b.AddParty(p1, a1, 0)
+	b.AddParty(p2, a2, 1)
+	return b
+}
+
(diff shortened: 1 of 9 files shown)


carpenterd777:
What's the difference between a `Party` and a `battleParty`?

dyc3:
A `Party` is simply a party of Pokemon. It is publicly available to users, and can easily be used for other things. A `battleParty` contains a `Party` and other things that a `Party` needs to be associated with in a `Battle`, like a team, pointers to active pokemon, and an agent. These things are _only_ relevant inside of a battle, so the separation is desirable.

util_test.go
```diff
@@ -374,3 +374,40 @@ func AlwaysRNG() *TestRNG {
...
+
+// Create a single Battle with only 1 pokemon in each party.
+// This function should only be used in tests.
+func New1v1Battle(p1 *Pokemon, a1 *Agent, p2 *Pokemon, a2 *Agent) *Battle {
+	return NewSingleBattle(NewOccupiedParty(p1), a1, NewOccupiedParty(p2), a2)
+}
+
+// Creates a new party to store Pokemon and assigns them to a team
+// This function should only be used in tests.
+func newBattlePartyOld(agent *Agent, team int) *battleParty {
```
carpenterd777, line 395: why is this function "old"? is it deprecated?

dyc3: Yeah, I'll mark it as such

carpenterd777 (approved):

PR #307, opened 2021-04-04
document what all the move flags mean

dyc3:


carpenterd777 (approved):
Some of these that aren't marked "non-gen 4" are also not applicable to Gen IV by nature of being related to a move or Ability from a later gen but overall good to know what these mean!

PR #321, opened 2021-04-06
Damage Dealt Utility

### LINKED ISSUE #282: HaveMore/HaveLess matchers
Instead of trying to determine a specific value of damage that is dealt based on some conditions that modify damage (stat stages, held items, moves, abilities, etc.), we can introduce a suite of matchers to check that some value is greater/less than another value. Gomega already supports `BeNumerically(comparator string, compareTo …interface{})`, but we can wrap this value to compare transactions.

The resulting API might look like this:
`Expect(t1).To(HaveMore("Damage", t2))` where `t1`, `t2` are DamageTransactions.

In simpler cases, we can also just check if the targetted Pokemon's HP is less/more after the round, but this won't work for things like accuracy.

adapap:
Adds `DamageDealt([]Transaction, *Pokemon) uint` and `battle.getTarget(int, int) target` helper methods. Started doing a test refactor, would like some input before I continue @dyc3. I am aiming to only change tests where checking for explicit targets is redundant (a.k.a. just comparing damage dealt).

closes #282

### DIFF
--- a/util_test.go
+++ b/util_test.go
@@ -343,7 +343,19 @@ func (matcher *orderedTransactionMatcher) NegatedFailureMessage(actual interface
 	)
 }
 
-// Tools for testing the library
+/* Tools for testing the library */
+// Check for damage dealt (if any) by a Pokemon in battle
+func DamageDealt(t []Transaction, p *Pokemon) uint {
+	for _, x := range t {
+		if v, ok := x.(DamageTransaction); !ok {
+			continue
+		} else if v.User == p {
+			return v.Damage
+		}
+	}
+	return 0
+}
+
 // Custom RNG struct which allows for predictable RNG output in a battle
 type TestRNG struct {
 	rolls  []bool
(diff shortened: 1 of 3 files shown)


util_test.go
```diff
@@ -343,7 +343,19 @@ func (matcher *orderedTransactionMatcher) NegatedFailureMessage(actual interface
 	)
 }
 
-// Tools for testing the library
+/* Tools for testing the library */
+// Check for damage dealt (if any) by a Pokemon in battle
+func DamageDealt(t []Transaction, p *Pokemon) uint {
```
dyc3, line 348: (nit) Let's make this return an int so we use `Equal(x)` instead of `BeEquivalentTo(x)` in the matchers

dyc3 (approved):

dyc3:
does this close #282 ?

adapap:
I think so. In my mind I had some idea where you pass in two battles and then compare the damage dealt with a matcher, but I think this is more appropriately handled in a setup/BeforeEach method, or potentially with a `battle.Reset()` method to easily compare two battles.

dyc3:
You could just do a deep copy of the battle to duplicate it

PR #334, opened 2021-04-10
Medicine Items

adapap:
Adding in all the medicine items


transactions.go
```diff
@@ -112,9 +112,135 @@ func (t ItemTransaction) Mutate(b *Battle) {
...
+	case ItemSodaPop:
+		b.QueueTransaction(HealTransaction{
+			Target: target,
+			Amount: 60,
+		})
+	case ItemSuperPotion:
+		b.QueueTransaction(HealTransaction{
+			Target: target,
+			Amount: 50,
+		})
```
dyc3, line 182: Since a lot of these healing items do the same thing, it would probably be better to put these in a `map` to look up the healing amount.

For `ItemMaxPotion` in particular, you could put `^uint(0)` for the max uint value.

dyc3 (changes requested):
Looks good mostly, just one concern

dyc3 (commented):
Looking much better. I think we could make it much better if we made the medicine data publicly accessible in `pokedex.go`, and if we could avoid remaking the map every time.

transactions.go
```diff
@@ -111,10 +111,88 @@ func (t ItemTransaction) Mutate(b *Battle) {
...
+		case ItemFullRestore, ItemMaxPotion, ItemMaxRevive, ItemRevivalHerb:
+			b.QueueTransaction(HealTransaction{
+				Target: target,
+				Amount: target.MaxHP(),
+			})
+		case ItemRevive:
+			b.QueueTransaction(HealTransaction{
+				Target: target,
+				Amount: target.MaxHP(),
+			})
```
dyc3, line 161: If you use `^uint(0)`, you can put these in `medicineData`

adapap: I think this is an inaccuracy because why wouldn't we just track the actual value instead? Also, I need the max HP anyway for revive and this actually pointed out that my test was wrong for revival items 👀

battle_item_test.go
```diff
@@ -0,0 +1,492 @@
...
+		return b, user
+	}
+
+	DescribeTable("Healing (HP)",
+		func(item Item, hp int) {
+			b, user := setup(item)
+			fainted := GeneratePokemon(PkmnIvysaur, defaultMoveOpt)
+			if item.Category() == ItemCategoryRevival {
+				user.CurrentHP = 0
+				hp = int(user.MaxHP())
```
dyc3, line 339: (nit) would prefer `expectedHp`

dyc3 (approved):

Issue #337, opened 2021-04-12
Recoil damage results in HealTransaction, which causes Health overflow and causes battle to never end


dyc3:
## Current Behavior
The battle described in the playground below does not end

**To Reproduce**
Code snippet to reproduce the problem (if applicable)
<details>
<summary>Playground</summary>

```go
package main

import (
	"fmt"
	"log"

	. "github.com/StevensSEC/pokemonbattlelib"
)

type smartAgent struct{}

func (smartAgent) Act(ctx *BattleContext) Turn {
	best := 0
	for i, target := range ctx.Opponents {
		if target.Pokemon.Defense() < ctx.Opponents[best].Pokemon.Defense() {
			best = i
		}
	}
	opponent := ctx.Opponents[best].Pokemon
	var strongestMove int
	var strongestDamage uint
	for i, move := range ctx.Pokemon.Moves {
		if move == nil {
			continue
		}
		damage := CalcMoveDamage(ctx.Battle.Weather, &ctx.Pokemon, opponent, move)
		if damage > strongestDamage {
			strongestDamage = damage
			strongestMove = i
		}
	}
	return FightTurn{
		Move:   strongestMove,
		Target: ctx.Opponents[best],
	}
}

func check(err error) {
	if err != nil {
		fmt.Printf("Error: %v", err)
	}
}

func main() {
	a1 := Agent(new(smartAgent))
	a2 := Agent(new(smartAgent))

	b := NewBattle()

	party0 := NewOccupiedParty(
		GeneratePokemon(410, WithLevel(52), WithMoves(74, 97, 88, 200), WithIVs([6]uint8{0x0, 0x0, 0x0, 0x0, 0x0, 0x0})),
		GeneratePokemon(144, WithLevel(49), WithMoves(254, 475, 344, 367), WithIVs([6]uint8{0x0, 0x0, 0x0, 0x0, 0x0, 0x0})),
		GeneratePokemon(137, WithLevel(51), WithMoves(18, 483, 360, 64), WithIVs([6]uint8{0x0, 0x0, 0x0, 0x0, 0x0, 0x0})),
		GeneratePokemon(60, WithLevel(52), WithMoves(229, 99, 310, 11), WithIVs([6]uint8{0x0, 0x0, 0x0, 0x0, 0x0, 0x0})),
	)
	party1 := NewOccupiedParty(
[...]

PR #339, opened 2021-04-12
using benchmark test to (kinda) fuzz for broken battles

dyc3:


benchmark_test.go
```diff
@@ -50,27 +52,72 @@ func randParty() *battleParty {
...
 				break
 			}
+			rounds++
+
+			if rounds > 100 {
+				dumpBattle(battle)
+				// log.Panic("Battle is stuck!")
+				stuck++
+				break
+			}
```
adapap, line 119: Should check if battle state has not changed for this IMO.

dyc3: That would not work if the state is oscillating or otherwise stuck in a loop. (example: #337)

adapap (approved):

PR #341, opened 2021-04-13
Plates / Type Enhancement Items

### LINKED ISSUE #340: Items: Plates/Type Enhancement
https://veekun.com/dex/items/misc

adapap:
Closes #340

Would like some form of #280 to be implemented for the type enhancement item tests.


dyc3 (commented):
Looks good to me, all it needs is tests

dyc3:
I think in order to do #280 effectively, we would have to separate battle logic into some `BattleRule` interface so we could enable/disable certain rules selectively. I don't think either of us have enough free time to do an overhaul that big right now. What exactly are you finding to be inconvenient for these tests? Maybe we could figure out something smaller that's more feasible right now.

adapap:
Actually yeah I can just make a move object directly. I completely forgot that we don't have to use `GetMove`. I needed to get moves of a specific type.

dyc3 (approved):

PR #344, opened 2021-04-14
http: quick and dirty hack for turn parsing

dyc3:


A-very-Cunning-ham (approved):
funky but if it works it works 👍

PR #346, opened 2021-04-16
improve misc docs, add some code examples

dyc3:


adapap (approved):
We should probably improve the README

PR #350, opened 2021-04-18
Implement most stat modifying moves

### LINKED ISSUE #331: Lots of stat changing moves that are already implemented are hardcoded, when they should use codegen data.
Currently, we have a few stat changing moves implemented here:

https://github.com/StevensSEC/pokemonbattlelib/blob/94be9db059024018d14820c4296e4b998acdb649/battle.go#L281-L284

Turns out, all of these stat changes are recorded here in `move_meta_stat_changes.csv`. **This is an easy win for a lot of functionality!**

dyc3:
- move tests
- implement most stat modifying moves

closes #331


battle_move_effects_test.go
```diff
@@ -10,38 +10,60 @@ var _ = Describe("Stat-changing moves", func() {
...
+		func(id MoveId, stat int) {
+			move := GetMove(id)
+			Expect(move.AffectedStat()).To(Equal(uint8(stat)))
 		},
-		Entry("Howl", MoveHowl),
-		Entry("Harden", MoveHarden),
+		Entry("Howl", MoveHowl, StatAtk),
+		Entry("Double Team", MoveDoubleTeam, StatEvasion),
+		Entry("Harden", MoveHarden, StatDef),
 	)
```
adapap: I think this check should just be moved to the other table so that you don't have to maintain two separate tables.

adapap (approved):
Minor suggestion for tests

battle_move_effects_test.go
dyc3: Valid, but this test is just meant to be a sanity check to make sure we are generating code for the data correctly. If this test fails, we know for sure that something is wrong with the code generation.

PR #352, opened 2021-04-19
Refactor `DamageTransaction` to be more similar to `HealTransaction`

dyc3:
- refactor move usage to new UseMoveTransaction
- move friendship/ev transactions to be created from FaintTransaction
- add UseMoveTransaction assertions to battle tests
- refactor orderedTransactionMatcher error message to make it easier to provide more debug info
- update playground template
- remove User from DamageTransaction

Ideally, I'd like to remove `Move` and `StatusCondition` from `DamageTransaction` as well, but I didn't want to completely remove the ability to determine the source of the damage without providing a replacement.

~~Would prefer #351 to be merged first~~


adapap:
We should probably have something like a `MoveContext` object. This we could use to keep track of whether the move was a critical hit, when a move occurred relative to opponents, what (if any) modifiers affected the damage, effectiveness, etc.

battle_move_test.go
```diff
@@ -23,3 +23,43 @@ func NewMove(power uint, category MoveCategory, moveType Type) MoveId {
 
 var TestMoveDefault = NewMove(10, MoveCategoryPhysical, TypeNormal)
 var TestMoveNoDamage = NewMove(0, MoveCategoryPhysical, TypeNormal)
+
+var _ = Describe("Move PP Consumption", func() {
+	a := Agent(new(dumbAgent))
+	It("should decrement move's PP by 1 when used", func() {
+		b := New1v1Battle(
+			GeneratePokemon(PkmnSquirtle, WithMoves(MoveSplash)), &a,
+			GeneratePokemon(PkmnSquirtle, WithMoves(MoveSplash)), &a,
```
adapap, line 32: Would prefer `TestMoveNoDamage`

battle_move_test.go
```diff
@@ -23,3 +23,43 @@ func NewMove(power uint, category MoveCategory, moveType Type) MoveId {
...
+		b.rng = AlwaysRNG()
+		Expect(b.Start()).To(Succeed())
+		t, _ := b.SimulateRound()
+		Expect(t).To(HaveTransactionsInOrder(
+			PPTransaction{
+				Amount: -1,
+			},
+			PPTransaction{
+				Amount: -1,
+			},
```
adapap, line 43: Don't we need a `*Move` to validate these?

adapap (changes requested):

battle_move_test.go
dyc3: Not really, since this is only testing to make sure the PP goes down when moves are used. It doesn't matter what the moves are.

adapap (approved):

PR #354, opened 2021-04-24
Generic Pokemon for Testing

adapap:
Goals:
- Create Pokemon with minimal side effects
- Simple reproduction and not verbose
- Extensible, such as with battle rules or other metadata


battle.go
```diff
@@ -23,6 +24,12 @@ type Battle struct {
 	results    BattleResults
 }
 
+type BattleRule int
+
+const (
+	BattleRuleNoFaint BattleRule = 1 << iota
+)
+
```
dyc3, line 32: We should define the default battle rules here as well, similar to the pokemon validation rules.
(nit, kinda) Also, would prefer all rules to be `1` if something is enabled, and `0` if something is disabled, so `BattleRuleNoFaint` -> `BattleRuleFaint`.

battle_move_test.go
```diff
@@ -9,10 +9,10 @@ package pokemonbattlelib
 // TODO: Move tests involving moves here
 
 // Creates a new move and returns the ID of the move created
-func NewMove(power uint, category MoveCategory, moveType Type) MoveId {
+func NewMove(power uint, moveType Type) MoveId {
```
dyc3, line 12: We should just make it take `MoveData`.

(nit) Would prefer `RegisterMove`

dyc3 (changes requested):

battle_move_test.go
```diff
@@ -9,20 +9,22 @@ import (
...
-	}
-	AllMoves = append(AllMoves, m)
+func RegisterMove(data MoveData) MoveId {
+	AllMoves = append(AllMoves, data)
 	return MoveId(len(AllMoves))
 }
 
-var TestMoveDefault = NewMove(10, MoveCategoryPhysical, TypeNormal)
-var TestMoveNoDamage = NewMove(0, MoveCategoryPhysical, TypeNormal)
+func RegisterMoveWithType(t Type) MoveId {
```
dyc3, line 17: This should be private, but remain in this file.

battle_move_test.go
```diff
@@ -9,20 +9,22 @@ import (
...
-func NewMove(power uint, category MoveCategory, moveType Type) MoveId {
-	m := MoveData{
-		Power:        power,
-		Category:     category,
-		Type:         moveType,
-		Accuracy:     100,
-		InitialMaxPP: ^uint8(0),
-	}
-	AllMoves = append(AllMoves, m)
+func RegisterMove(data MoveData) MoveId {
```
dyc3, line 12: This should go in pokedex.go

transactions.go
```diff
@@ -600,6 +600,11 @@ type FaintTransaction struct {
 }
 
 func (t FaintTransaction) Mutate(b *Battle) {
+	if b.ruleset&BattleRuleFaint == 0 {
+		pkmn := b.getPokemon(t.Target)
```
dyc3, line 604: Log a message to `blog` here that mentions that fainting is disabled, and the pokemon has been restored to max heath.

dyc3 (changes requested):
Just a couple of small changes

dyc3 (approved):

PR #357, opened 2021-04-28
Switch Turn

### LINKED ISSUE #36: Implement SwitchTurn type and add case to SimulateRound()
This will represent a turn where a Trainer changes their active Pokemon (the one currently battling) to a Pokemon in their party that has not yet fainted.

See: https://github.com/StevensSEC/pokemonbattlelib/wiki/Requirements#pokemon-switching-out

Switching Pokemon always takes precedence to any `FightTurn`s that may occur, essentially placing it in the highest priority bracket (higher than any move would have). (The exception is the move Pursuit, but this has not been implemented yet. It should be kept in mind while designing the type though.)

- [x] Switching Pokemon passes the turn
- [x] Switching Pokemon takes place before any `FightTurn`s
- [x] A Trainer cannot switch to a fainted Pokemon
- [x] A Trainer cannot switch to the already active Pokemon
- [x] Add case to `SimulateRound()`

adapap:
Closes #36


battle.go
```diff
@@ -277,6 +281,19 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
 				Item:   t.Item,
 				Move:   t.Target.Pokemon.Moves[t.Move],
 			})
+		case SwitchTurn:
+			party := b.GetParty(&t.Target)
+			if _, ok := party.activePokemon[t.Target.partySlot]; ok {
+				blog.Panic(ErrorCannotSwitch)
```
dyc3, line 287: We _really_ should stop panicking to throw errors. Definitely a job for another PR though.

dyc3 (commented):
Looking good so far

battle.go
```diff
@@ -315,3 +320,13 @@ type ItemTurn struct {
 func (turn ItemTurn) Priority() int {
 	return 1
 }
+
+// A turn to represent switching an active Pokemon for a different, inactive Pokemon in battle.
+type SwitchTurn struct {
+	Current AgentTarget // The current active target being swapped out
```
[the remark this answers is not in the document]
adapap: Yes that's true, will change that.

battle.go
```diff
@@ -87,11 +87,16 @@ func (b *Battle) getPokemon(t target) *Pokemon {
...
+	pokemon := party.pokemon()
+	if t.slot >= uint(len(pokemon)) {
 		panic(ErrorPartyIndex)
 	}
-	return p[t.slot]
+	slot := t.slot
+	if party.IsActivePokemon(t.slot) {
+		slot = party.activePokemon[t.slot]
+	}
+	return pokemon[slot]
```
dyc3, line 99: There needs to be a comment or something to explain the logic here

battle_round.go
```diff
@@ -135,6 +135,17 @@ func (b *Battle) SimulateRound() ([]Transaction, bool) {
...
+			party := b.GetParty(turn.User)
+			if !party.IsActivePokemon(turn.User.slot) {
+				blog.Panic(ErrorCannotSwitch)
+			}
+			pkmn := b.getPokemon(t.Target.target)
+			if pkmn.CurrentHP == 0 {
+				blog.Panic(ErrorCannotSwitch)
+			}
+			party.SetInactive(turn.User.slot)
+			party.activePokemon[turn.User.slot] = t.Target.slot
```
dyc3, line 148: We should probably make a distinction between active pokemon slots and party slots. This confused me the first time I looked at it.

dyc3 (approved):

PR #361, opened 2021-04-30
update comparePokemon to compare the pointer instead, because we can do that now

dyc3:


util_test.go
```diff
@@ -18,10 +18,7 @@ func comparePokemon(a, b *Pokemon) bool {
 	} else if a == nil || b == nil {
 		return false
 	}
```
adapap, line 20: Is this case necessary now that we have the `a == b` fallthrough?

adapap (approved):

util_test.go
dyc3: Nope, removing it

PR #366, opened 2021-05-01
implement status inflicting moves

dyc3:


battle_move_test.go
```diff
@@ -19,6 +19,78 @@ func registerMoveWithType(t Type) MoveId {
...
 var TestMoveNoDamage = RegisterMove(MoveData{Name: "No Damage", InitialMaxPP: 100})
+var TestMoveInflictBurn = RegisterMove(MoveData{Name: "Inflict Burn", Category: MoveCategoryStatus, InitialMaxPP: 100, Ailment: StatusBurn})
+var TestMoveDamageAndInflictBurn = RegisterMove(MoveData{Name: "Damage and Inflict Burn", Category: MoveCategoryPhysical, InitialMaxPP: 100, AilmentChance: 100, Ailment: StatusBurn, Power: 10})
+
+var _ = Describe("Move Status Inflict", func() {
+	a := Agent(new(dumbAgent))
+	It("should inflict burn", func() {
+		b := New1v1Battle(
+			GeneratePokemon(PkmnSquirtle, WithMoves(TestMoveInflictBurn)), &a,
+			GeneratePokemon(PkmnSquirtle, WithMoves(TestMoveNoDamage)), &a,
```
adapap, line 30: Same as other PR, `PkmnWithMoves(TestMoveInflictBurn)`

battle_move_test.go
```diff
@@ -19,6 +19,78 @@ func registerMoveWithType(t Type) MoveId {
...
+		)
+		b.rng = AlwaysRNG()
```
adapap, line 32: ~Not sure if any RNG is necessary here, because moves with 0 accuracy will not miss.~
Forgot that status effects are RNG...

battle_move_test.go
```diff
@@ -19,6 +19,78 @@ func registerMoveWithType(t Type) MoveId {
...
+		t, _ := b.SimulateRound()
+		Expect(t).To(HaveTransaction(
+			InflictStatusTransaction{
+				Target:       b.getPokemonInBattle(1, 0),
+				StatusEffect: StatusBurn,
+			},
+		))
+	})
+
+	It("should deal damage and inflict burn", func() {
```
adapap, line 43: This test seems redundant to me. We have moves that test damage and moves that test burns + inflicting burn. This would imply we need to do this for every move with side-effects manually and I don't see it as necessary.

adapap (changes requested):
Just changes to tests. Implementation looks good

battle_move_test.go
dyc3: Moves that are purely status inflicting moves are in the meta category `ailment`. All `AilmentChance` for those moves are zero, and they are `MoveCategoryStatus`.

Moves that are in the meta category `damage+ailment`, are **not** `MoveCategoryStatus`, and `AilmentChance > 0`.

I wasn't able to cleanly combine these 2 cases without accidentally printing out `Unimplemented status move` to the battle log, so they are separate in the code. Besides, this distinction would warrant having both of these cases anyway.

adapap (approved):

PR #367, opened 2021-05-01
refactor all tests to use PkmnWithMoves instead

dyc3:
Generated with:

```
gofmt -r 'GeneratePokemon(b, WithMoves(a)) -> PkmnWithMoves(a)' -w *.go
``﻿`

I'm not sure if we want to do this right now because we need to change `PkmnWithMoves` and others to use a new `RegisterPokemon` function that works how `RegisterMove` works.


adapap:
The main reason right now for register instead is for debugging purposes, so I don't think it matters.

dyc3:
`RegisterPokemon` is still a public API that I would like to have anyway. Mainly made this PR as a demonstration of how `gofmt` can be used to refactor code.

dyc3:
Stale

PR #368, opened 2021-05-01
RFC: API Change for Targets

### LINKED ISSUE #362: Refactor: Targets & Pokemon Pointers
New API will use only Target{Party, Slot} as `[2]int` keys for pretty much every operation. This will replace any existing calls to `battle.getPokemon`, `battle.getPokemonInBattle`, `battle.getTarget`, `target`, and related fields in the current API. The idea is to not worry about the specific pointers as too many factors potentially cause it to be invalidated/changed. Agents will still receive a copy of the Pokemon for reference?

adapap:
### _*Not working on refactoring tests until there is a consensus on API changes_

This PR proposes a new API based around the fact that `target` structs will be passed around, which are simply a composite key of (`party`, `slot`) of an active? Pokemon. The reasons for this proposal include:

- Pointer/memory management is difficult
- Easier to validate turns and transactions
- Hashable key
- Less data passed around to/from client

Potential/only? drawback:
- Must retrieve Pokemon pointer from battle as-needed
  - This is not an expensive operation, as it is a constant-time composite key lookup.
This has the side effect of always ensuring that Pokemon have updated data, including if they are switched out, fainted, etc.

### Summary of API changes
- Moved pre, during, and post-round logic to a new file: `battle_round.go`
- `target` is now defined as `struct{party: uint, slot: uint}` (negative indices will not occur!)
- All transactions use `target` objects instead of pointers
- party.activePokemon is now a `map[uint]uint` instead of a map to pointers to easily implement `SwitchTurn` in the future
- BattleContext is reworked to lazily compute values: `Self()`, `Allies()`, `Opponents()`, `Targets()`

Removed the following helper methods as they are redundant, replaced, or otherwise disposable:
- `battle.getPokemonInBattle`
- `battle.getTarget`
- `battle.GetTargetsRef`
- `battle.getContext`, `battle.GetRoundContext` renamed to `battle.GetBattleContext`
- `battleParty.GetActivePokemon` (this one might be needed still or replaced)

Closes #362


dyc3:
It would be a good idea to at least refactor the test in `agent_test` to make sure you're making deep copies.

adapap:
@dyc3 Modified a chunk of tests, and the example in benchmark, to use the new API. Not much actually has to change on the implementation end - the benefits are mostly internal organization.

I want to double check that the copy is actually a copy passed to agents if you could verify that.

dyc3 (approved):
Looks good to me

PR #370, opened 2021-05-03
[http] add locks around map writes, fixes #369

### LINKED ISSUE #369: fatal error: concurrent map writes
## Current Behavior
While stress testing my ssw project, I encountered this race condition. It occurs in the code below:

```go
func (hb *httpBattle) QueueNextTurn(targetId int, turn Turn) {
	hb.queuedTurns[targetId] = turn
}
```

Should be easily fixed with a lock.

**To Reproduce**
Queue multiple turns for the same battle at the same time.

## Expected Behavior
Should not crash.

**Evironment:**
 - OS: linux

**Additional context**

dyc3:


A-very-Cunning-ham (approved):
fix looks good to me

PR #371, opened 2021-05-04
Add gender randomization, and some additional codegen

dyc3:
- add gender randomization by default when generating pokemon
- add gender info to codegen
- add more species data to codegen
- add move meta category to code gen


pokemon.go
```diff
@@ -39,6 +40,8 @@ type PokemonData struct {
 	BaseStats  [6]int // base stats of a Pokemon
 	EvYield    [6]int // effort points gained when Pokemon is defeated
 	GrowthRate int
+	IsBiGender bool  // true if this Pokemon must have `Male` or `Female`, false if this Pokemon must be `Genderless`
```
adapap: Why not `HasGender` or `IsGenderless`?

pokemon_test.go
```diff
@@ -210,3 +210,21 @@ var _ = Describe("Pokemon Data", func() {
...
 	})
 })
+
+var _ = Describe("RandomGender", func() {
+	DescribeTable("Should always return one gender",
+		func(rate int, gender Gender) {
+			for i := 0; i < 10000; i++ {
+				Expect(RandomGender(uint8(rate))).To(Equal(gender))
+			}
+		},
```
adapap, line 220: This feels like a convoluted way to test this functionality. At this point we're just testing if a RNG is working. We can probably do just fine with 2-3 hard-coded iterations.

adapap (approved):
Some thoughts

pokemon_test.go
dyc3: This is equivalent of hard coding 10000 iterations in a row. This guarantees that `RandomGender` holds the behavior we are expecting.

pokemon.go
dyc3: It makes Pokemon genderless by default, which is our previous behavior.

PR #378, opened 2021-05-14
Struggle

### LINKED ISSUE #329: Struggle when out of PP for all moves
blocked #322

adapap:
Work in progress

Closes #329

### DIFF
--- a/transactions.go
+++ b/transactions.go
@@ -16,6 +16,20 @@ type UseMoveTransaction struct {
 
 func (t UseMoveTransaction) Mutate(b *Battle) {
 	user := b.getPokemon(t.User)
+	// Struggle conditions
+	if t.Move.CurrentPP == 0 {
+		// Ensure that struggle is forced (TODO: move to FightTurn validation)
+		for _, m := range user.Moves {
+			if m != nil && m.CurrentPP > 0 {
+				panic("cannot use a move with 0 PP")
+			}
+		}
+		b.QueueTransaction(UseMoveTransaction{
+			User:   t.User,
+			Target: t.Target,
+			Move:   GetMove(MoveStruggle),
+		})
+	}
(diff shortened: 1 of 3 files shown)


dyc3:
We also need to have a pokemon validation rule that disallows pokemon from having struggle.

transactions.go
```diff
@@ -16,6 +16,20 @@ type UseMoveTransaction struct {
 
 func (t UseMoveTransaction) Mutate(b *Battle) {
 	user := b.getPokemon(t.User)
+	// Struggle conditions
+	if t.Move.CurrentPP == 0 {
+		// Ensure that struggle is forced (TODO: move to FightTurn validation)
+		for _, m := range user.Moves {
+			if m != nil && m.CurrentPP > 0 {
+				panic("cannot use a move with 0 PP")
```
dyc3, line 24: Should panic with an error instead of a string

dyc3 (changes requested):
Needs to have a new `PokemonValidationRule` to ensure that you can't battle with pokemon that know `MoveStruggle` as one of their 4 moves.

https://github.com/StevensSEC/pokemonbattlelib/blob/5ad3be35b9b35a4e1b8f5f75e668ad60c91a6ca3/pokemon.go#L420

Otherwise, looks good.

adapap:
> Needs to have a new `PokemonValidationRule` to ensure that you can't battle with pokemon that know `MoveStruggle` as one of their 4 moves.

To clarify, are you saying that struggle should not be allowed as one of the Pokemon's moves when validating a Pokemon? Wouldn't this be better to include with learnsets?

dyc3:
Yeah, that's correct. We don't have learnsets yet, so it'll be a temporary thing we can replace when we do learnsets.

dyc3 (approved):

Issue #379, opened 2021-05-15
Draco meteor should decrease the user's special attack by 2.


dyc3:
## Current Behavior
Draco meteor should decrease the user's special attack by 2.

https://bulbapedia.bulbagarden.net/wiki/Draco_Meteor_(move)

**Additional context**

PR #380, opened 2021-05-16
Document move meta categories, and add a test for MoveMetaCategoryDamageRaise

dyc3:
- document all move meta categories
- add a test for MoveMetaCategoryDamageRaise (#379)

This does NOT close the issue #379. It adds a pending test that is skipped, but can be enabled when #379 is implemented.

### DIFF
--- a/move.go
+++ b/move.go
@@ -51,20 +51,20 @@ const (
 type MoveMetaCategory uint8
 
 const (
-	MoveMetaCategoryDamage MoveMetaCategory = iota
-	MoveMetaCategoryAilment
-	MoveMetaCategoryNetGoodStats
-	MoveMetaCategoryHeal
-	MoveMetaCategoryDamageAilment
-	MoveMetaCategorySwagger
-	MoveMetaCategoryDamageLower
-	MoveMetaCategoryDamageRaise
-	MoveMetaCategoryDamageHeal
-	MoveMetaCategoryOhko // One hit knock out
-	MoveMetaCategoryWholeFieldEffect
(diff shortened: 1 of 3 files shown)


move.go
```diff
@@ -51,20 +51,20 @@ const (
...
+	MoveMetaCategoryHeal                                     // Heals the user or a target ally.
+	MoveMetaCategoryDamageAilment                            // Deal damage to target pokemon, and have a chance to apply a status condition to the target of the move.
+	MoveMetaCategorySwagger                                  // Raise the target's stats, and apply confusion.
+	MoveMetaCategoryDamageLower                              // Deal damage to target pokemon, and have a chance to apply a stat modifier to the target of the move.
+	MoveMetaCategoryDamageRaise                              // Deal damage to target pokemon, and have a chance to apply a stat modifier to the user of the move.
+	MoveMetaCategoryDamageHeal                               // Deal damage to target pokemon, and restore some of the user's HP. However, this is effectively a noop, because this functionality is only present on moves that have `Drain != 0`.
+	MoveMetaCategoryOhko                                     // One hit knock out. If the move hits the target pokemon, it is generally guarenteed to die (there are exceptions). Always fails if the target pokemon's level is greater than the user's level.
```
adapap, line 63:
```suggestion
	MoveMetaCategoryOhko                                     // One hit knock out. If the move hits the target pokemon, it is generally guaranteed to die (there are exceptions). Always fails if the target pokemon's level is greater than the user's level.
```

adapap (approved):
Typo, otherwise LGTM

PR #388, opened 2021-09-01
add iron fist ability

dyc3:


battle_abilities_test.go
```diff
@@ -24,3 +24,24 @@ var _ = Describe("Scrappy", func() {
...
+			GeneratePokemon(PkmnBidoof, WithLevel(15), WithMoves(MoveCometPunch)), &a,
+		)
+		b.rng = AlwaysRNG()
+		Expect(b.Start()).To(Succeed())
+		t, _ := b.SimulateRound()
+		normalDamage := DamageDealt(t, target{1, 0})
+		boostedDamage := DamageDealt(t, target{0, 0})
+		Expect(normalDamage).To(BeNumerically(">", 0))
+		Expect(boostedDamage).To(BeNumerically(">", 0))
+		Expect(boostedDamage).To(BeNumerically(">", normalDamage))
```
adapap, line 43: 2nd line here implies the first one is true.

battle_abilities_test.go
```diff
@@ -24,3 +24,24 @@ var _ = Describe("Scrappy", func() {
...
+
+var _ = Describe("Ability: Iron Fist", func() {
+	a := Agent(new(dumbAgent))
+
+	It("should have punching moves deal 1.2x damage", func() {
+		b := New1v1Battle(
+			GeneratePokemon(PkmnMachop, WithLevel(15), WithAbility(AbilityIronFist), WithMoves(MoveCometPunch)), &a,
+			GeneratePokemon(PkmnBidoof, WithLevel(15), WithMoves(MoveCometPunch)), &a,
+		)
+		b.rng = AlwaysRNG()
```
adapap, line 36: OOC, why not `SimpleRNG()`?

adapap (approved):
LGTM!

battle_abilities_test.go
dyc3: So the attacks don't miss

battle_abilities_test.go
dyc3: yeah but this gives a little bit more granularity for what went wrong

battle_abilities_test.go
adapap: The goal of simple RNG was to have a factory that produces attacks that always hit but don't have other side effects (stat changes, crits, etc.). Maybe it is too difficult now to trace this behavior?

PR #389, opened 2021-09-02
add battle types, implement double battles

dyc3:


battle.go
```diff
@@ -152,13 +184,27 @@ func (b *Battle) Start() error {
...
+		parties := b.GetPartiesOnTeam(i)
+		pkmnPerParty := b.Type.GetMaxActivePokemon() / uint(len(parties))
+		for _, party := range parties {
+			toSendOut := pkmnPerParty
+			if toSendOut > uint(len(party.pokemon())) {
+				toSendOut = uint(len(party.pokemon()))
+			}
+			for j := uint(0); j < toSendOut; j++ {
+				party.SetActive(j)
+			}
```
adapap, line 206: A bit confused on the logic here (also, what is the casting to `uint` for?). If you have team 1 have 4/2 pokemon respectively, doesn't that mean that one party doesn't have pokemon for a while and the other sends out two Pokemon?   Or is the assumption that the parties are just combined?

Another thought - is there validation somewhere that the sum of the parties' Pokemon is no more than 6 for a double battle?

adapap (approved):
Approve but would like comment addressed

battle.go
dyc3: Casting because pkmnPerParty is `uint`, can't compare types that don't match.

> A bit confused on the logic here

This is to handle all the cases:
- Team has 2 parties, so each party should send out 1 pokemon each. (`pkmnPerParty == 1`)
- Team has 1 party, so the party must send out 2 pokemon. (`pkmnPerParty == 2`)
- Team has 1 party, but only 1 pokemon, so the party can only send out 1 pokemon. (`pkmnPerParty == 2`, but `toSendOut` becomes 1)

IIRC, on a team of 2 parties, if one party runs out of pokemon, the other party does not send out another pokemon. They are limited to 1 active pokemon per party. Regardless, should be pretty easy to fix if I'm wrong anyway.

> is there validation somewhere that the sum of the parties' Pokemon is no more than 6 for a double battle?

No because only parties are restricted to 6 pokemon, not teams. A team could have 2 parties of 6 pokemon each, and that would be totally valid.

PR #391, opened 2021-09-02
feat: ability air lock

adapap:


battle_round.go
```diff
@@ -15,6 +17,9 @@ func (b *Battle) preRound() {
 				StatusEffect: StatusSleep,
 			})
 		}
+		if pkmn.Ability == AbilityAirLock {
+			b.metadata[MetaWeatherDisabled] = true
+		}
```
dyc3, line 22: Should this be another transaction so it's possible to announce it? Something like `AbilityEffectTransaction`.

battle_abilities_test.go
```diff
@@ -6,7 +6,30 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("Scrappy", func() {
+var _ = Describe("Ability: Air Lock", func() {
+	a := Agent(new(dumbAgent))
+
+	It("should negate the effects of weather", func() {
```
dyc3, line 12: Can we have a test to see if the effect gets removed when the pokemon is removed from the battlefield?

dyc3 (changes requested):

battle_round.go
adapap: I think it would be hard to make a transaction generic enough to be able to handle these types of cases. Right now the metadata fields are the only `interface{}` types. I also considered this being a potential battle rule (to disable weather), which doesn't make sense for it being a transaction.

This seems like the easiest way to ensure that we always have this check in the beginning of a round.

